PR #6, opened 2025-06-26
docs: Setup readthedocs for medcat

alhendrickson:


medcat-v2/docs/main.md
```diff
@@ -1,4 +1,4 @@
-# Medical  <img src="https://github.com/CogStack/cogstack-nlp/blob/main/medcat-v2/media/cat-logo.png?raw=true" width=45>oncept Annotation Tool (v2)
+# Medical  <img src="https://github.com/CogStack/cogstack-nlp/blob/main/medca/media/cat-logo.png?raw=true" width=45>oncept Annotation Tool (v2)
```
mart-r, line 1: Typo here - should be without `/medca`.

medcat-v2/README.md
```diff
@@ -1,10 +1,10 @@
-# Medical  <img src="https://raw.githubusercontent.com/CogStack/MedCAT/main/media/cat-logo.png" width=45> oncept Annotation Tool (version 2)
+# Medical  <img src="https://github.com/CogStack/cogstack-nlp/blob/main/medca/media/cat-logo.png?raw=true" width=45> oncept Annotation Tool (version 2)
```
mart-r, line 1: Type here - should be without `/medca`

mart-r (changes requested):
Other than the 2 typos, I think this looks good.

PS: Didn't realise you were already doing this. So #7 will probably need to be dropped. This is a probably a better approach anyway - having the media central rather than on a per service basis.

medcat-v2/docs/main.md
alhendrickson: Thanks!

mart-r (approved):
Look good to me

PR #7, opened 2025-06-26
CU-8699j87ej Fix further links (and add logo)

mart-r:
Some furthe link fixes that were pointing to old stuff.

Also added the logo to medcat-v2 to avoid pointing to the login in the v1 repo.


mart-r:
Looks like it was all addressed in #6

PR #8, opened 2025-06-26
build: Add UV and Pip caching

alhendrickson:
MedCAT 2 Test build improved from 20 minutes to 10 minutes.

Using default caching described in https://docs.astral.sh/uv/guides/integration/github/#caching.

Added mandatory uv.lock file, it's good practice anyway. One todo is add a local dev guide, unless I'm missing it.


mart-r:
Just as a note, all but the 3.9 workflows already ran in around 12 minutes. e.g:
https://github.com/CogStack/cogstack-nlp/actions/runs/15902962856/job/44850224953

But still, getting 3.9 down to the same level is definitely a win!

mart-r:
I guess my point was that I wasn't quite sure what the target of it was. I.e - like you said - it doesn't get propagated to the end user in any way (and we wouldn't want it to).

And the other concern was the fear of it becoming another thing to maintain. Though that's certainly a solveable problem, either through a pre-commit hook or a workflow, or perhaps even something else.
For some reason I'd lean towards having a workflow that checks the `uv.lock` to make sure it doesn't need upgrading and it would then fall for the dev to implement the change (though this would likely be more or less automated anyway).

With that said, I'm not sure the lock file is necessarily too useful at dev time since (as far as I know) it's not portable on different python versions or OSs or CPU architecture, or with and without GPU (`cuda`) stuff. Though definitely useful for debugging complex issues to ensure compatible environments.

So it sounds like we'll keep it for:
- Caching
- Consistency in terms of CI

And I think we should plan a for a way to ensure it's up to date in one way or the other.

PR #11, opened 2025-06-26
Update release tag

mart-r:
This will use `medcat-v<major>.<minor>.<patch>` for release tag instead of `medcat/v<major>.<minor>.<patch>`.

Fix on the workflow step as well as on the release scripts.


mart-r:
Not going this route

PR #19, opened 2025-06-30
build: Only run actions based on path changed filter

alhendrickson:


mart-r:
Looks like it didn't run / finish all the jobs. I tried to restart the failed jobs, but only 1 restarted.

Are we hitting the concurrent jobs limit here? Though that shouldn't affect rerunning only a portion.

.github/workflows/medcat-v2-tutorials_main.yml
```diff
@@ -4,8 +4,11 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
-
+    paths:
+      - 'medcat-v2/**'
```
mart-r, line 8: Just as a comment, I don't think this _currently_ has much effect since the tutorials target a hard coded version rather than the one in the PR. But in principle that is a separate issue we should address.

.github/workflows/medcat-service_docker.yml
```diff
@@ -2,10 +2,13 @@ name: medcat-service - Build Images
 
 on:
   push:
-    branches:
-      - '**'
+    branches: [ main ]
     tags:
-      - 'v*.*.*'  # e.g., v0.1.1
+      - 'medcat-service/v*.*.*'  # e.g., medcat-serice/v0.1.1
+  pull_request:
```
mart-r, line 8: Just commenting here with the changes.
Does this actually make sense? This workflow pushes to docker hub. I don't think it makes sense to push to docker hub upon each PR, does it?
@alhendrickson

alhendrickson: In my view - yes I want to push every time to docker hub. There's no storage limit at all, and it is very fast due to the layering and caching, so I dont see a real reason not to. To further justify: in the example docker provide for their official  github action, it pushes to docker every push to git, dockerhub really dont care how much you want to send them. Practical benefits are this is now true e2e, it pulls and runs the exact image from dockerhub for this branch for a test

As the side note - I would still prefer to run every action every single push, the only reason I see not to right now is the concurrency limit in github combined with the long action times right now. As a culture here I'd want to move towards us not caring if the runners do "unnecessary work", until there's some real reason, and I think a lot of these actions including other repos have a bit of premature optimisation going on in terms of when they run, instead of how fast they run

mart-r: The problem is this means changes are pushed before a review of the PR.
While I'm pretty sure we have workflow disabled for outside PRs (so no direct attack vector), once you enable it, any PR will immediately push the changes to docker hub upon workflow. So even if every other workflow fails, the image in docker hub was already pushed as part of this workflow.

IMHO this should only run on new commits on the `main` branch rather than on PRs.

PR #24, opened 2025-07-02
CU-8699my5eg Add install bundles to releases

mart-r:
This PR aims to add install bundles to releases.

The install bundles it will (hopefully) include:
- All target 64bit x86 and a Linux environment
- Includes common optional extras
  - `rel-cat`
  - `meta-cat`
  - `deid`
  - `spacy`
- There's 2 bundle for each supported python version
  - One that's CPU only (torch-wise)
  - ~~One that allows GPU as well (torch-wise)~~ (couldn't make it work as it was too big)
  - Supported versions are (currently) 3.9, 3.10, 3.11, and 3.12

So TLDR, there will be ~~8~~ 4 total install bundles included.

The README for the install bundles can be found here:
https://github.com/CogStack/cogstack-nlp/blob/CU-8699my5eg-add-release-bundles/medcat-v2/.release/install_bundle_readme.md

PS:
~~Currently, the workflow will run during a PR, but that will (somewhat obviously) need to be changed before merging~~
Removed / fixed

EDIT:
Had to remove the GPU install bundle because it's too big. The limit is 2GiB and the cuda-enabled bundle is around 3GB.


mart-r:
Just a comment:

Had to remove the GPU install bundle because it's too big. The limit is 2GiB and the cuda-enabled bundle is around 3GB.

alhendrickson (approved):

alhendrickson:
Commenting this so I can remember it, out of our slack talk 

> 
> If we target this flow that you've put down, targeting just the air gap user in a trust - I'm feeling like it is a good specific problem to solve.
> "download it from GitHub, copy it to their one drive (which is available inside and outside), download it to their VPN machines, and install based on that. "
> 
> I'm seeing two main features:
> 
> * Making a zip containing everything, that can go into one drive
> * Make a script that can pip install from that zip, even if it's just for an example
> 
> Today, you've got the CPU zip in github, which looks good to me. Install script is basically what you do for the smoketest I think, but will be good to document this somewhere. I'm totally on board with what you have for this now
> 
> And these assumptions:
> * There's some manual actions on by the users side, which we can't automate - we cant control their one drive
> * Site unblocking and hosting not really required, its done by users 
> * Whether the zip already exists, or users have to make it by running a script, doesnt affect it really
> * There can be some client side scripts done by the user
> 
> 
> So in future I'm thinking we could do stuff like this:
> * Script making the zip by the user, instead of making it for them in github
> * Make scripts with options for GPU, architectures etc. Doesnt have to just be 10 complete zip files, but could still be
> * Add more requirements on the user side? EG put all this into docker images, they `docker run` locally instead which would solve OS issues, file size use, etc. . Mount a local drive in docker, run our image, which magically creates a zip
>

PR #26, opened 2025-07-03
CU-8699nbgbh Test against changes

mart-r:
This PR aims to test tutorials and demo work app against changes in the current PR rather than the version pre-defined in the files.

Stuff done
- [x] Tutorials (verified installed branch-changed in workflow logs)
- [x] Demo web app (same as above)
- [x] Add support for medcat service


.ci/patch_notebooks_and_requirements.py
```diff
@@ -0,0 +1,71 @@
+import sys
+import pathlib
+import re
+from functools import partial
+
+
+repo_url = "git+https://github.com/CogStack/cogstack-nlp"
```
alhendrickson, line 7: Hey this looks good, and its definitely going to work and its going to be great testing everything in one go - this is the dream :D

My feeling is that there's got to be an easier way than this though? It seems like a common use case, my gut reaction is regexing and mutating files is going to get hard to maintain later. Long term view is maybe we will need to bring proper repo tooling in like bazel, but anyway...


As a few thoughts:

- Can it specify the local folder directly instead of github branch, essentially `pip install ../medcat-v2/...`? Would then be able to repeat locally and not need github

- If editing the file is the way to go, can we change to make it create temp copies instead of editing ? EG change this to create a "build/ci-requirements.txt" or whatever instead of editing the requirments.txt file. Keeping it immutable should help make it clear later, and help locally

- Instead of editing, can I override the version instead of editing requirements.txt? `pip install -r requirements.txt some-arg`. Or in general some native way to make it really predefined here. Just feels like there should be some built in feature for this over editing the file

- Lastly - I do see the need for this regex and flow with the ipynb - but maybe we can just do one method for the demo app/service, and another for the notebooks? Like the requirements.txt version can basically be "remove medcat line from requirements.txt, pip install medcat@my-version`, and save the regex for the ipynb. Reason is to keep it simple where it can be, over trying to reuse the complex one

mart-r:
The latest push from today also includes stuff to run the 2 medcat-service workflows upon changes to `medcat-v2` folder and run them against the current state. So that's both the tests as well as the docker image build and hub push (along with the integration tests).

.github/workflows/medcat-service_docker.yml
```diff
@@ -51,17 +55,22 @@ jobs:
             # Create version tag based on tag prefix
             type=match,pattern=medcat-service/v(\d+\.\d+\.\d+),group=1
 
+      - name: Make medact-v2 available within build
+        run: cp -r ../medcat-v2 medcat-v2
```
alhendrickson, line 59: Minor for later - thinking about the docker caching and image size here

Do you know how much of the medcat-v2 folder is actually needed? As this will put everything into docker, like the docs and whatever too, which invalidates the cache and rebuilds. 

Can we be more selective with the copy? Or alternatively do a ton of work in the .dockerignore but that feels harder

```
cp -r ../medcat-v2/medcat medcat-v2/medcat
cp -r ../medcat-v2/toml-file? medcat-v2/whatever

```

medcat-service/Dockerfile
```diff
@@ -28,5 +28,12 @@ RUN for spacy_model in $SPACY_MODELS; do python -m spacy download $spacy_model;
 # Copy the remaining files
 COPY . /cat
 
+# NOTE: for test/CI time, this will install medcat based on local path
+ARG REINSTALL_CORE_FROM_LOCAL=false
```
alhendrickson, line 32: Minor for later - again for the caching 

Following the usual copy requirements.txt pattern - can we change this to copy what is needed, then copy the py files?

Eg

```
# Install medcat-v2
COPY ./medcat-v2 /cat/wherever

# NOTE: for test/CI time, this will install medcat based on local path
ARG REINSTALL_CORE_FROM_LOCAL=false
# pip install ./medcat-v2
...
# Copy the remaining files
COPY . /cat
```

.github/workflows/medcat-service_docker.yml
mart-r: I think what's _needed_ is the source folder, the `.toml` and the license. Though I'd need to double-check.
But certainly not docs or stuff like that.

medcat-service/Dockerfile
mart-r: The reason I've done it the way I'm doing it now is that if I copy the folder during the workflow, if someone wants to build outside the workflow, they wouldn't be bloated with the entire context of the `medcat-v2` folder built into the image.

But given that the source (+ other small bits) is probably quite small (order of megabytes, I'd guess), i we only took what we needed, that'd probably be negligible in the grand scheme of things.

PR #27, opened 2025-07-04
CU-8699nk284 Update MedCAT service to v2

mart-r:
Update MedCAT service to run on MedCAT v2.

Currently just hard coded version.


medcat-service/Dockerfile
```diff
@@ -6,6 +6,9 @@ ENV CRYPTOGRAPHY_DONT_BUILD_RUST=1
 WORKDIR /cat
 COPY ./requirements.txt /cat
 
+# NOTE: need git for URL based installs
+RUN apt-get update && apt-get install -y git
```
alhendrickson, line 10: For a later improvement, I guess we can switch to the multi stage builds like you made for other things. This at least is highlighting the build dependency vs runtime anyway. Though I get here it's going to change like 0.01% of the total image size...

alhendrickson (approved):
I'm surprised it's working with the v1 pack I've put in, but great that it is! I'd look to bundle a v2 pack now in there as well for the testing

medcat-service/Dockerfile
mart-r: Well, in the long (or medium) run, we should be able to remove this. Once we're installing based on PyPI instead of a GH link, we don't need `git` anymore.

PR #41, opened 2025-07-10
feat(medcat-service): Create liveness and readiness APIs for use by K8s

alhendrickson:


medcat-service/medcat_service/nlp_processor/medcat_processor.py
```diff
@@ -84,7 +92,34 @@ def __init__(self):
...
+
+    def _check_medcat_readiness(self) -> bool:
+        readiness_text = "MedCAT is ready and can get_entities"
+        try:
+            result = self.cat.get_entities(readiness_text)
+            self.log.debug("Result of readiness check is" + str(result))
+            self.log.info("MedCAT processor is ready")
+            return True
+        except Exception as e:
+            self.log.error(f"MedCAT processor is not ready. Failed the readiness check: {e}")
```
mart-r, line 105: If we pass the error to the logger through `exc_info=e` keyword argument, we can log the full stack trace instead of just the message (which the current implementation would show).

mart-r (approved):
Slight nag on the logged error.
Other than that, looks good to me.

mart-r (approved):
Not sure why the force push (so don't know the exact diffs between my last review and now), but looks good.

PR #47, opened 2025-07-14
CU-8699td0xq: Move to v2 model pack

mart-r:
Convert v1 model pack to v2 format and use that instead.

EDIT:
I thought there were other bits in the repo that had v1 model packs, but couldn't really find anything else. `medcat-v2` does have old model packs, but they should (mostly) be for legacy conversion, and - as such - should remain.

### DIFF
--- a/.github/workflows/medcat-service_docker.yml
+++ b/.github/workflows/medcat-service_docker.yml
@@ -112,6 +112,9 @@ jobs:
   integration_test:
     runs-on: ubuntu-latest
     needs: build
+    strategy:
+      matrix:
+        variant: [v1, v2]
     steps:
       - name: Check out code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


alhendrickson:
Can we leave the v1 model pack in as well? From a test point of view - if supporting v1 is still part of the functionality of the service, I'd want to leave it in so it's tested (and it's pretty easy to leave it in...). From a selfish point of view, it means I don't have to do any work on my side.

If you'd rather not have v1 in there - let's rename it to just "example-model-pack.zip" and remove the v2 designation, basically to make the config easier going ahead.

For details - right now this change from the file name -v1 to -v2 is actually a breaking change for the work I'm doing (as in - if I repull my :latest service now in k8s, it breaks as my env var wont work). It's obviously fine for me to update the var, but in general I'd want to either avoid making the breaking change, or at least make it easier to change going ahead by making the file name more generic

PR #51, opened 2025-07-17
refactor(medcat-service): Add pydantic. Add types to Health and Info APIs

alhendrickson:
Initial work moving towards a FastAPI rewrite which wants types for the APIs.  Next step is to add to the remaining APIs

Plan is to add these types to know the exact input/output of each API, then afterwards actually change the impl with safety.

### DIFF
--- a/medcat-service/medcat_service/types.py
+++ b/medcat-service/medcat_service/types.py
@@ -3,8 +3,6 @@ from typing import List, Literal, Optional, Union
 
 from pydantic import BaseModel
 
-BaseModel.model_config['protected_namespaces'] = ()
-
 
 class HealthCheckResponse(BaseModel):
     """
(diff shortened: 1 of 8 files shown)


medcat-service/medcat_service/types.py
```diff
@@ -1,9 +1,33 @@
-from typing import Literal, TypedDict
+from datetime import datetime
+from typing import List, Literal, Optional, Union
 
+from pydantic import BaseModel
 
-class HealthCheckResponse(TypedDict):
+BaseModel.model_config['protected_namespaces'] = ()
```
mart-r, line 6: The way it's done here is a global change. It would affect every other `pydantic` model that inherits from `BaseModel` (and doesn't explicitly set protected namespaces), including ones in dependencies.
I don't think that's what we want.

I think the intent here was probably for this to affect all the models in this module? I would just define a base model `class NoProtectedBaseModel(BaseModle, protected_namespaces=())` (e.g https://docs.pydantic.dev/2.0/usage/model_config/#:~:text=class%20Model(BaseModel%2C%20extra%3D%27forbid%27)%3A) and inherit from that where applicable.

mart-r (changes requested):
I don't think the global change to the config makes sense.
Other than that, looks good.

PR #58, opened 2025-07-23
CU-8699xagnz bug(medcat-service): Add test showing that meta_anns_filters in payload errors out

alhendrickson:
[2025-07-23 11:53:47,069] [ERROR] API: Traceback (most recent call last):
  File "MedCATservice/medcat_service/api/api.py", line 51, in process
    result = nlp_service.nlp.process_content(payload['content'], meta_anns_filters=meta_anns_filters)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/MedCATservice/medcat_service/nlp_processor/medcat_processor.py", line 160, in process_content
    all(e['meta_anns'][task]['value'] in filter_values
  File "MedCATservice/medcat_service/nlp_processor/medcat_processor.py", line 160, in <genexpr>
    all(e['meta_anns'][task]['value'] in filter_values
        ~~~~~~~~~~~~~~^^^^^^
KeyError: 'Presence'


alhendrickson:
> This actually fails at the linting check instead. So it doesn't actually get to your added test.
> 
> Probably easiest to just change the relevant line(s) to:
> 
> ```python
>         self.assertEqual(
>             response.status_code, 500,
>             "Bug: KeyError: 'Presence' on all(e['meta_anns'][task]['value'] in filter_values")
> ```

Right, yes it does, I'll fix.

Note the test should "pass" to show that 500 is returned, but really we know it should be 200.

The reason I made the test was to check whether the feature was broken, or if I broke it in the move to fastapi.

I tried this test against the version in the old repo as well, and confirm the feature seems to error out.

mart-r:
Is this just a PR to introduce a test that's passing but shouldn't? To me that feels a little backwards. I'd expect tests to describe / test intended behaviour.
If we've identified an issue, shouldn't we introduce a test that fails instead? And then work on a fix? Though that'd be TDD and we don't generally tend to do that.
Or am I missing something?

mart-r:
Yeah, that makes sense. Don't want to throw away work you did.

But in my experience, you don't really look at tests unless they fail. So since this test is not failing, unless we explicitly mark this down as an issue elsewhere (granted you may have done that somewhere), this test will keep running (successfully) without anyone checking whether it should pass or not. Only if/when the a user tries to use the underlying feature does it become evident that there is a test, but the test just confirms that things don't work as they should. Though at that point, that's at least some information you use to get back to this PR.

I do think this comes down to opinion. I think I would generally just document the issue on somewhere (potentially with a test that would fails) and omit the PR.

I more or less wanted to understand the rationale behind a test that confirms faulty behaviour. But I suppose the trail back to the PR with more context is probably sufficient documentation to make this better than just an issue logged somewhere else.

alhendrickson:
Right - this has also been tracked in clickup under CU-8699xagnz, and in the PR name / commit. The issue is in backlog, it's not closed, so there's nothing lost here. I'm definitely still logging issues. 

So I feel like - given the choice between having an issue in clickup backlog with no test, compared to having the same issue that also has a test, the one with the test is just more information? When someone comes to work on it and modifies the code to fix the bug, this test will then fail with a clear message like "this is a bug", and can confidently correct the test.

Anyway, as it's an interesting philosophical thing I have been thinking about today - 

I wonder if its down to what we think unit testing does. IMO they test "expected" behaviour, which isn't necessarily "intended" (though should be...). For example, I can make tests around some 3rd party library I'm using to validate my usage of it, but obviously I don't have any real idea what the library "intended" to do, like maybe it's real purpose is stealing my data. I black box testing is probably the name.

I think the tests that actually confirm if things are behaving as they are meant to would be acceptance tests or BDD etc, not quite the same level as these ones here. EG it would be possible to have 100% coverage of medcat service in a unit test that all say it's behaving as expected, but actually the service it provides "annotating vets animal notes" or whatever hits 0% of the intended purpose.

PR #62, opened 2025-07-31
CU-8699zymxk: fix v1 build workflow

mart-r:
This should hopefully fix the v1 build workflow that otherwise always fails during pushes merges into main branch.


alhendrickson:
Do you have any ideas on how this magically failed one day without any changes to the folder?

From my view - I didn't change anything in medcat-v1, though the build failed on me

mart-r:
Pretty sure it'll have been `transformers` pushing a change.
I.e this was released on the same day:
https://pypi.org/project/transformers/4.54.1/
https://github.com/huggingface/transformers/releases/tag/4.54.1
And this a few days earlier:
https://pypi.org/project/transformers/4.54.0/
https://github.com/huggingface/transformers/releases/tag/v4.54.0

One of them must have either changed the argument name or removed the `*kwargs` option, or both.

PR #63, opened 2025-07-31
build(medcat-service): Fix latest tag in docker getting set on a release

alhendrickson:
https://github.com/CogStack/cogstack-nlp/actions/runs/16372680383/job/46264753505

This build triggered on a release of medcat, and updated the latest tag in docker making it go back in time vs main

Two fixes here:
- Stop it building on a release, I think a tag is enough
- Stop it setting :latest tag when built on a tag (the flavour setting)

### DIFF
--- a/.github/workflows/medcat-service_docker.yml
+++ b/.github/workflows/medcat-service_docker.yml
@@ -10,8 +10,6 @@ on:
       - 'medcat-v2/**'
       - 'medcat-service/**'
       - '.github/workflows/medcat-service**'
-  release:
-    types: [published]
 
   # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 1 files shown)


mart-r (approved):
I guess this issue only arises because the releases are based on release branches which aren't (necessarily) kept in sync with the `main` branch of the entire repo.
I think that does make sense overall, since the release is drafted against changes in the `medcat-v2` folder. So changes in other folders shouldn't affect it.

So as long as we don't do releases that are specific to the service, I think this change makes a lot of sense. If we end up doing that, we'd need to build some logic for this. But a problem for another day.
And if we're tagging based on release tag anyway, we're not loosing anything.

PR #67, opened 2025-08-06
Adding functionality for offline loading

shubham-s-agarwal:
This change modifies the BERT model loading logic to support offline initialization from local files only, removing any dependency on online resources.

This change modifies the BERT model loading logic to support offline initialization from local files only, removing any dependency on online resources.

New behavior:

- If `load_bert_pretrained_weights=True`: load pretrained BERT weights (requires internet access).
- If `load_bert_pretrained_weights=False`: initialize BERT with random weights

Note: we always recommend loading the model weights from the provided model packs. In case user does not want to load the model pack, then using `load_bert_pretrained_weights`=False is unadvisable, however in our scripts, we always load the model packs.

- If `saved_dir_path` is provided: Load BERT config from local file in the specified path.
- If `saved_dir_path`  is not provided: Fallback to downloading config from HuggingFace.

### DIFF
--- a/medcat-v1/medcat/config_meta_cat.py
+++ b/medcat-v1/medcat/config_meta_cat.py
@@ -133,10 +133,6 @@ class Model(MixingConfig, BaseModel):
 
     NB! For these changes to take effect, the pipe would need to be recreated.
     """
-    load_bert_pretrained_weights: bool = False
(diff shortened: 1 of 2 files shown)


mart-r (approved):
Looks good, thanks!

PR #74, opened 2025-08-11
Add package dir to v1 release / production workflow

mart-r:
This should allow for a v1 (patch) release.

Needed for patch release due to #67 .


mart-r:
The workflow failure is quite frustrating here. The workflow (or install targets) haven't changed since it was last successful (3 days ago, https://github.com/CogStack/cogstack-nlp/actions/runs/16829058972/job/47672154440).

I _think_ this has to do with `setuptool-scm` updates. They've pushed 9.1.0 an hour ago. Yet they have 4 yanked versions in the last week.
But so far, my efforts to force the use of the older one have been futile.

PR #95, opened 2025-08-13
chore: Move all Medcat V1 projects into a subfolder

alhendrickson:


mart-r (commented):
If it's in the `v1` folder, do we want to perhaps remove teh `-v1` suffix of the folders? Seems to be duplicating the information.

There's a bunch of links in the `v1` folder that are either still pointing at old stuff (e.g `v1/medcat-v1-tutorials/README.md` has a bunch of links; the tutorial notebooks refer to raw files within the old /separate repo; the v1 docs and readme refer to old/separate tutorials repo; the `v1/medcat-v1/docs/main.md` refers to a non-existing `https://github.com/CogStack/cogstack-nlp/tree/main/medcat-trainer/`; )
But I'm not sure if it's worth trying to fix all that...

PR #97, opened 2025-08-13
chore(medcat-trainer): CU-869a4br6j Create a copy of the v1 medcat-trainer in the v1 folder

alhendrickson:
…ainer in the /v1 folder


mart-r (approved):
We may want to make a note of the CodeQL issues for the future.

But since they're already in the other one anyway, I don't see it as a priority.

PR #98, opened 2025-08-14
bug(medcat-service): Fix DeID mode fails with PydanticSerializationError

alhendrickson:
Used the solution from https://github.com/CogStack/cogstack-nlp/blob/e55bf6cd3fd453c0feafb10f50d33d5732c7ace0/medcat-demo-app/webapp/demo/views.py

I didnt work out why it was returning np.float for DeID mode but not for the regular mode (noting regular just returns 1 in the tests, so could be that). I also didnt work out a better way to make pydantic convert them automatically, seemed like a lot. 

```
INFO:API:Process_result is text='[**********] says hello' annotations=[{0: {'pretty_name': 'PATIENT', 'cui': 'PATIENT', 'type_ids': [], 'source_value': 'John smith', 'detected_name': '', 'acc': np.float32(0.77323574), 'context_similarity': np.float32(0.77323574), 'start': 0, 'end': 10, 'id': 0, 'meta_anns': {}, 'context_left': [], 'context_center': [], 'context_right': []}}] success=True timestamp='2025-08-14T14:06:14.668+00:00' elapsed_time=0.076584845 footer=None
ERROR:API:Unable to process data
  File "/home/ali/workspace/github.com/alhendrickson/CogStack/cogstack-nlp/medcat-service/medcat_service/routers/process.py", line 55, in process
    log.info("Response is " + response.model_dump_json())
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ali/workspace/github.com/alhendrickson/CogStack/cogstack-nlp/medcat-service/.venv/lib/python3.12/site-packages/pydantic/main.py", line 441, in model_dump_json
    return self.__pydantic_serializer__.to_json(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'numpy.float32'>
```

### DIFF
--- a/medcat-service/medcat_service/nlp_processor/medcat_processor.py
+++ b/medcat-service/medcat_service/nlp_processor/medcat_processor.py
@@ -5,6 +5,7 @@ import os
 import time
 from datetime import datetime, timezone
 
+import numpy as np
 import simplejson as json
 from medcat.cat import CAT
 from medcat.cdb import CDB
@@ -121,6 +122,8 @@ class MedCatProcessor():
             if "entities" in entities.keys():
                 entities = entities["entities"]
 
+            self._fix_floats(entities)
+
             if self.entity_output_mode == "list":
                 entities = list(entities.values())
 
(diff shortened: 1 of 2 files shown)


mart-r:
I wasn't able to dig up more than what you already found... It's annoying though... I would have thout `numpy` would have fixed this on their end. But perhaps that's wishful thinking.

PR #99, opened 2025-08-14
CU-869a4wpek Trainer qa workflow fix

mart-r:
Attempting to fix the trainer QA workflow.

I _think_ this was just an issue of it trying to push the same version (1.0.0) over and over again. And that was just failing because the test PyPI rejected the duplicate.

EDIT:
Also made the workflow run at PR time so I can check without a merge.


tomolopolis (approved):
lgtm

PR #104, opened 2025-08-20
bug(medcat-service): DeID bulk processing fix 

alhendrickson:
Bugs fixed:

- deid_multi_texts was renamed in medcat2 so missing here
- Bulk Process for DeID passed the wrong data format to deid_multi_text
- Bulk process flow never appeared to read the result of deid_multi_texts, instead it just returned the input with no deid

Things to do in a later PR:
- Rename back to deid_multi_texts
- Make deid_multi_text also return the annotations
- Make service bulk process flow more explicit

### DIFF
--- a/medcat-service/medcat_service/test/test_deid.py
+++ b/medcat-service/medcat_service/test/test_deid.py
@@ -81,7 +81,7 @@ class TestMedcatServiceDeId(unittest.TestCase):
             0,
             "CU-869a6wc6z No annotations are currently returned by the bulk API",
         )
-        
+
         # Note: CU-869a6wc6z commended out these asserts until annations are returned
         # ann = actual["result"][0]["annotations"][0]["0"]
         # self.assertEqual(ann["pretty_name"], expected["pretty_name"])


medcat-service/medcat_service/nlp_processor/medcat_processor.py
```diff
@@ -209,13 +209,13 @@ def process_content_bulk(self, content):
...
 
         try:
+            text_input = MedCatProcessor._generate_input_doc(content, invalid_doc_ids)
             if self.DEID_MODE:
-                # TODO 2025-07-21: deid_multi_texts doesnt exist in medcat 2?
-                ann_res = self.cat.deid_multi_texts(MedCatProcessor._generate_input_doc(content, invalid_doc_ids),
-                                                    redact=self.DEID_REDACT)
+                text_to_deid_from_tuple = (x[1] for x in text_input)
+
+                ann_res = self.cat.deid_multi_text(list(text_to_deid_from_tuple),
```
alhendrickson, line 216: To highlight this one - medcat v1 had deid_multi_texts, v2 has deid_multi_text

If not deliberate, I'd probably name it back with the plural to match the non-deid one.

Would suggest to do in a later PR if we want to

mart-r: Thanks, will do!

alhendrickson:
> Some minor gripes. But looks good to me overall!

Cheers - I've fixed all these

PR #111, opened 2025-08-21
CU-869a2kpv0 Add method for model card load off disk

mart-r:
Adds method to load model card off disk without loading model pack (though this may include unzipping the archive).
Also adds a few simple tests for it.
And updates the test time model pack to a newer one (the old one was in an older format that didn't have the model card).

### DIFF
--- a/medcat-v2/medcat/cat.py
+++ b/medcat-v2/medcat/cat.py
@@ -9,7 +9,6 @@ from contextlib import contextmanager
 
 import shutil
 import zipfile
-import tempfile
 import logging
 
 from medcat.utils.defaults import DEFAULT_PACK_NAME, COMPONENTS_FOLDER


medcat-v2/medcat/cat.py
```diff
@@ -834,6 +834,43 @@ def get_model_card(self, as_dict: bool = False) -> Union[str, ModelCard]:
...
+        Args:
+            model_pack_path (str): The path to the model pack (zip or folder).
+            as_dict (bool): Whether to return as dict. Defaults to False.
+
+        Returns:
+            Union[str, ModelCard]: The model card.
+        """
+        # unpack if needed
+        if model_pack_path.endswith(".zip"):
+            model_pack_path = cls.attempt_unpack(model_pack_path)
```
alhendrickson, line 864: Hey as I've also got some interest in this

Instead of unpacking the whole zip, can you just pull the json out direct? 

EG I have a 1GB zip but just need 10kb json right now, would prefer not doubling the storage usage in the cloud just to show model info

mart-r: Should be able to do that. My _assumption_ was that if you're trying to access it, it's fine to unpack as well. But I'm sure I don't need to tell you what I do when I assume...

mart-r: Added an option to do that, but disbaled by default.

alhendrickson: Yeah it's fair, it's a "nice to have one day, maybe never" ask from my side, probably bigger from the mlops side

For me - I want to deploy like 1000 instances of the service, but maybe I'm not sure exactly which instance has the model I want, so I could check the /info API first 1k times to find the right one. Right now from an service perspective, medcat is lazy loaded on the first process call, so would cool if /info didnt use any resources up.

I have no plan to really do this but in general it feels like it should be able to get info with minimal side effects.

medcat-v2/medcat/cat.py
```diff
@@ -834,6 +836,59 @@ def get_model_card(self, as_dict: bool = False) -> Union[str, ModelCard]:
             return model_card
         return json.dumps(model_card, indent=2, sort_keys=False)
 
+    @overload
```
tomolopolis, line 839: purpose of `@overload` here rather than `-> Union[str, ModelCard]` as you've done in the impl?

tomolopolis (approved):
great thanks - lgtm!

PR #112, opened 2025-08-22
CU-869a7mjaa: Add simplified method of getting pipe from CAT object

mart-r:
Otherwise ended up using protected attribute in many places.


tomolopolis (approved):
lgtm - ❤️  this

mart-r:
> lgtm - ❤️ this

Too many emojis - clearly ChatGPT!

PR #114, opened 2025-08-26
CU-869a3ycrj Trusted publisher for PyPI

mart-r:
This PR adds trusted publisher based releases to PyPI:
- [x] medcat-trainer QA to TestPyPI
- [ ] medcat-trainer full release to PyPI

PS:
Currently makes the workflow run on pushes to this branch to test it out, this will be removed.


mart-r:
NOTE: The publish to TestPyPI was successful in the last commit:
https://github.com/CogStack/cogstack-nlp/actions/runs/17236413273/job/48902264415?pr=114

PR #117, opened 2025-08-27
chore(medcattrainer): update to latest release

tomolopolis:
latest release up on dev, was release tagged a while ago...


mart-r (approved):
Looks good!
Though to note - this is a change to the v2-specific trainer. The v1-version on here still says `v2.22.1` everywhere.

PR #118, opened 2025-08-27
chore(medcat): CU-869a98zwq: use old name

tomolopolis:
- when converting a dir of models, use the old name by default so folks know what model is what.  
- update readme to link tutorial and provide a quick ref directly in the README.


mart-r (approved):
Yeah - that makes sense - preserve the name!

Looks good to me.

PR #129, opened 2025-09-05
feat(medcat-trainer): CU-869adnv5w Add Postgres database support to trainer

alhendrickson:
This is effectively a requirement for a kubernetes deployment, as soon as more than one instance gets run (eg on a rolling deployment)

### DIFF
--- a/medcat-trainer/docker-compose-example-postgres.yml
+++ b/medcat-trainer/docker-compose-example-postgres.yml
@@ -3,7 +3,7 @@
 services:
   # medattrainer services
   medcattrainer:
-    image: cogstacksystems/medcat-trainer:v2.23.1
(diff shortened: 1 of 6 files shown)


medcat-trainer/docker-compose-example-postgres.yml
```diff
@@ -0,0 +1,83 @@
+# Default compose yml file - uses latest build of MedCATtrainer services. Default passwords and example
+# projects are not used.
+services:
+  # medattrainer services
+  medcattrainer:
+    image: cogstacksystems/medcat-trainer:latest
```
alhendrickson, line 6: (Note this yml technically doesnt work yet. I did test this with local builds)

tomolopolis (approved):
lgtm

PR #138, opened 2025-09-15
Bug(medcat): CU-869ag0tqj Fix regex tokenizer dashes

### LINKED ISSUE #128: Words after hyphens are removed by CDBMaker
pls see repost below

mart-r:
The regex based tokenizer would treat all words that had some punctuation in them as skip words and would skip them.
And because of that, these tokens would be ignored since the tagger would tag them as skip words because they contained punctuation.

This PR fixes that by separating the punctuation in the regex based tokenizer example.

This should fix #128.


medcat-v2/tests/model_creation/test_cdb_maker.py
```diff
@@ -29,6 +33,39 @@ def setUpClass(cls):
...
+        cls.maker.prepare_csvs([cls.cui_df, ], full_build=True)
+
+    def test_has_cui(self):
+        self.assertIn(self.cui, self.cdb.cui2info)
+
+    def test_has_full_names(self):
+        sep = self.config.general.separator
+        for _name in self.namelist:
+            # lowercase and pre-condition
+            name = _name.lower()
```
alhendrickson, line 55: Minor, I'd personally prefer just explicitly writing out what the expected values are, mainly for the self documenting and there's just 3 of them. `"wernicke~korsakoff~syndrome"` (if that is it), I guess the fact that I'm not sure is kind of the reason why.

alhendrickson (approved):
lgtm, now it's much simpler with your latest commit

PR #141, opened 2025-09-17
CU-869ah00vt Fix memory leak (mc-service)

vladd-bit:
Bug: when processing docs for a long amount of time memory usage begins to increase > 7-8gb usage for de-id models after processing a 20-30k after 3+ hours, this bug is reproducible for normal annotation models too.

Fix: try to limit the lru cache to maximum 1 instance for the processor and other objs.

### DIFF
--- a/medcat-service/medcat_service/main.py
+++ b/medcat-service/medcat_service/main.py
@@ -7,7 +7,7 @@ from fastapi.responses import JSONResponse
 
 from medcat_service.config import Settings
 from medcat_service.demo.gradio_demo import io
-from medcat_service.dependencies import set_global_processor
+from medcat_service.dependencies import set_global_processor, set_global_settings
 from medcat_service.nlp_processor.medcat_processor import MedCatProcessor
 from medcat_service.routers import admin, health, process
 from medcat_service.types import HealthCheckFailedException
@@ -30,8 +30,6 @@ async def lifespan(app: FastAPI):
         medcat = MedCatProcessor(settings)
         app.state.medcat = medcat
 
-    app.state.settings = settings
-    app.state.medcat = medcat
     app.state.title = "MedCAT Service",
     app.state.summary = "MedCAT Service",
     app.state.contact = {
@@ -45,6 +43,7 @@ async def lifespan(app: FastAPI):
     },
     app.state.root_path = settings.app_root_path
 
+    set_global_settings(settings)
     set_global_processor(medcat)
     log.debug("MedCAT Service lifespan setup complete")
 


mart-r (changes requested):
IMHO this doesn't fix the underlying issue.

We expect to have 1 `Settings` instance. And as such, we expect to have 1 `MedCATProcessor` instance.
But what this PR does is allow for multiple instances of `MedCATProcessor` while only caching one at a time.
Not only does this mean that there could still be multiple instances around (if they're still in use), but it also means that each one must have a slightly different `Settings` instance (because if they were identical they should hash to the same thing and you should still get the same processor).
Now why you'd have multiple different `Settings` instances is a little bizarre to me. Because the only reason I can think of is if `os.environ` gets manually changed.

PS:
Open to discussion on leaving this in the sate you've proposed. But I'm just a little concerned regarding the underlying issue(s).

medcat-service/medcat_service/dependencies.py
```diff
@@ -10,14 +10,14 @@
...
-@lru_cache
+@lru_cache(maxsize=1)
 def get_settings() -> Settings:
     settings = Settings()
     log.debug("Using settings: %s", settings)
     return settings
 
 
-@lru_cache
+@lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
vladd-bit: this was just the easy way to enforce the mcat proc singleton,  we can switch to the lifespan manager then and see how it behaves

medcat-service/medcat_service/dependencies.py
```diff
@@ -10,14 +10,14 @@
 log = logging.getLogger(__name__)
 
 
-@lru_cache
+@lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
alhendrickson: This makes sense to me, feels right to make it explicitly a global singleton. Would keep that log line in from before and really confirm it never gets created again

medcat-service/medcat_service/main.py
```diff
@@ -1,28 +1,56 @@
...
+async def lifespan(app: FastAPI):
+
+    log = logging.getLogger(__name__)
+    log.debug("Starting MedCAT Service lifespan setup")
+
+    # allow overriding settings and medcat processor for testing
+    settings = getattr(app.state, "settings", None)
+    if settings is None:
+        settings = Settings()
+        app.state.settings = settings
```
mart-r, line 26: This is duplicated on line 33. Should be able to remove this line here and just keep the one after as far as I can tell.

medcat-service/medcat_service/main.py
```diff
@@ -1,28 +1,56 @@
...
+
+    medcat = getattr(app.state, "medcat", None)
+    if medcat is None:
+        medcat = MedCatProcessor(settings)
+        app.state.medcat = medcat
```
mart-r, line 31: This is duplicated on line 34. Should be able to remove this line here and just keep the one after as far as I can tell.

medcat-service/medcat_service/main.py
```diff
@@ -1,28 +1,56 @@
...
-    license_info={
+    app.state.license_info = {
         "name": "Apache 2.0",
         "identifier": "Apache-2.0",
     },
-    root_path=settings.app_root_path,
-)
+    app.state.root_path = settings.app_root_path
+
+    set_global_processor(medcat)
```
mart-r, line 48: What about `set_global_settings`? Does that never get set?
I guess `get_global_settings` is never called either, so maybe we don't need that at all?

medcat-service/medcat_service/dependencies.py
```diff
@@ -1,26 +1,52 @@
...
+    _settings = request.app.state.settings
+    log.debug("Using settings: %s", _settings)
+    return _settings
 
-@lru_cache
-def get_medcat_processor(settings: Annotated[Settings, Depends(get_settings)]) -> MedCatProcessor:
-    log.debug("Creating new Medcat Processsor using settings: %s", settings)
-    return MedCatProcessor(settings)
 
+def set_global_settings(settings: Settings) -> None:
```
mart-r, line 21: Is this needed? Doesn't seem to be called anywhere?

medcat-service/medcat_service/dependencies.py
```diff
@@ -1,26 +1,52 @@
...
-from fastapi import Depends
+from fastapi import Depends, Request
 
 from medcat_service.config import Settings
 from medcat_service.nlp_processor.medcat_processor import MedCatProcessor
 
 log = logging.getLogger(__name__)
 
+processor_singleton: Optional[MedCatProcessor] = None
+settings_singleton: Optional[Settings] = None
```
mart-r, line 12: Is this needed? The relevant methods (`get_global_settings` and `set_global_settings` don't seem to be called anywhere.

mart-r (changes requested):
A few duplicate lines.
And a few questions of relevance for the global settings.

But I think this does look better to me!

medcat-service/medcat_service/main.py
```diff
@@ -1,28 +1,55 @@
...
 from medcat_service.types import HealthCheckFailedException
 
-settings = get_settings()
 
-app = FastAPI(
-    title="MedCAT Service",
-    summary="MedCAT Service",
-    contact={
+@asynccontextmanager
+async def lifespan(app: FastAPI):
```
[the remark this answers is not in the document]
alhendrickson: To clarify - I mean, taking a step back, is it not possible to keep it all basically how it was before, just with the settings being explicitly global?

dependencies.py

```python
_def_settings: Optional[Settings] = None

def get_settings() -> Settings:
    global _def_settings
    if _def_settings is None:
        _def_settings = Settings()
    return settings
...
@lru_cache
def get_medcat_processor(settings: Annotated[Settings, Depends(get_settings)]) -> MedCatProcessor:
    log.debug("Creating new Medcat Processsor using settings: %s", settings)
    return MedCatProcessor(settings)
```

Then there are no changes needed anywhere else (I think?)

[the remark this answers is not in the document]
mart-r: We probably want this merged in at some point sooner rather than later.

I think we can all agree that the main point of this PR was to avoid the memory leak from and that this PR seems to accomplish that.

It sounds like @alhendrickson is concerned as to why the previous - seemingly standard - approach caused issues. And this may be worth looking into. But I think @vladd-bit 's point is that he opted for a different fix - one that gives more control over lifecycle management. So a fix + move forward.

So given the above, I think we should just move forward with and merge this PR. It does seem to fix the issue. And it seems to move forward in terms of lifecycle management.

Does that sound good to all parties?

alhendrickson:
Hey for this one -

I've put the gunicorn config into it's own PR that hopefully easily goes in https://github.com/CogStack/cogstack-nlp/pull/201 

I've then got another PR that adds just the global medcat processor on here https://github.com/CogStack/cogstack-nlp/pull/200/files. The issue I was really fixing was a threading issue (service breaks in k8s under load), but I think it is the same underlying problem - let me know if it's possible to test it for the memory leak too.

PR #146, opened 2025-09-23
Refactor(medcat):CU-869ak0v7n Add type hints to util methods

mart-r:
Add type hints to some (mostly) MetaCAT related utility methods.

This will help downstream parts easier use of these parts and makes it easier to maintain and test against.


tomolopolis (approved):
lgtm

PR #147, opened 2025-09-25
feat(medcat): CU-869aknekf Add mapping to ontologies

mart-r:
This PR adds back the functionality from v1 that added mappings to other ontologies to the output of `get_entities`.

This PR adds this back, but with a slightly differtent approach. I've opted to make this a config option that can be changed if / when needed, instead of an extra argument to the method(s).

### DIFF
--- a/medcat-v2/medcat/config/config.py
+++ b/medcat-v2/medcat/config/config.py
@@ -252,11 +252,6 @@ class General(SerialisableBaseModel):
     map_cui_to_group: bool = False
     """If the cdb.addl_info['cui2group'] is provided and this option enabled,
     each CUI will be mapped to the group"""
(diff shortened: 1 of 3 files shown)


baixiac (approved):
Thanks and LGTM. Minor comment added.

medcat-v2/medcat/config/config.py
```diff
@@ -252,11 +252,17 @@ class General(SerialisableBaseModel):
     map_cui_to_group: bool = False
     """If the cdb.addl_info['cui2group'] is provided and this option enabled,
     each CUI will be mapped to the group"""
-    simple_hash: bool = False
```
tomolopolis, line 255: this config prop is now removed?

PR #150, opened 2025-09-30
Upload export docs

tomolopolis:
currently broken since introducing new CDB / Vocab or ModelPack requirements on a Project.

Adds another method on the client to allow easy uploading of a list of exported projects. Tutorial Notebook also provided.


mart-r (approved):
Looks good to me.

Though this would necessitate a client release as well (currently there's nothing but a dev release anyway).

PR #151, opened 2025-09-30
Trainer upload export test

tomolopolis:


mart-r (approved):
More tests = more better!

PR #153, opened 2025-09-30
CU-869anj8ub: Add initial commit of WWC conversion

mart-r:
This is the `working_with_cogstack` replacement.

It will use `medcat_den` under the hood for centralised model storage.


tomolopolis:
Task linked: [CU-869anj8ub Move medcat-scripts to cogstack-nlp](https://app.clickup.com/t/869anj8ub)

alhendrickson:
Hey can you add some detail on why some of the files were left out?

EG I see that 1_create_model and 2_train_model aren't included.

One guess is that these are both completely covered in the tutorials as well.

PR #157, opened 2025-09-30
Trainer update load examples

tomolopolis:
update the load-examples script that is used on startup for a basic mcv2 model pack and update to a small M-IV extract


tomolopolis:
thanks - yeah not sure why the change from #152  is included here. That's already been released and is on pypi

PR #159, opened 2025-10-01
fix for config.general.show_nested_entities

pisong314:
**The issue**
whether the config.general.show_nested_entities flag is True or False, there seems to be no effect to the output.

**Example**
if the text is “chest pain” which matches 1) “chest pain” greedily 2) “chest” 3) “pain”, it seems to produce all 3 entities at the end at all time regardless of the flag.

**What's wrong**
- the current logic in postprocessing.py doesn't seem to be filtering anything.
- the filtering logic doesn't appear to look at the flag at all.

**The fix**
- re-implementation of the filtering logic to simply use token index to filter.
- the flag is passed on and checked during post-processing.


pisong314:
I just added  # type: ignore  from the original code back.  seems to build ok now.

PR #160, opened 2025-10-01
feat(medcat): CU-869apb8ju Better ontology mapping

mart-r:
This PR improves ontology mapping.

### DIFF
--- a/medcat-v2/medcat/cat.py
+++ b/medcat-v2/medcat/cat.py
@@ -535,15 +535,8 @@ class CAT(AbstractSerialisable):
         other_onts = self.config.general.map_to_other_ontologies
         if other_onts:
             if other_onts == "auto":
-                self.config.general.map_to_other_ontologies = other_onts = [
-                    key.removeprefix("cui2")
-                    for key in self.cdb.addl_info
-                    if key.startswith("cui2")
-                ]
-                logger.info(
-                    "Automatically finding ontologies to map to: %s",
-                    other_onts)
-            for ont in self.config.general.map_to_other_ontologies:
+                other_onts = self._set_and_get_mapped_ontologies()
+            for ont in other_onts:
                 if ont in out_dict:
                     logger.warning(
                         "Trying to map to ontology '%s', but it already "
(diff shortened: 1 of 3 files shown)


medcat-v2/medcat/config/config.py
```diff
@@ -252,13 +252,17 @@ class General(SerialisableBaseModel):
...
     each CUI will be mapped to the group"""
-    map_to_other_ontologies: list[str] = ["opcs4", "icd10"]
+    map_to_other_ontologies: Union[Literal["auto"], list[str]] = "auto"
     """Which other ontologies to map to if possible.
 
     This will force medcat to include mapping for other ontologies in
     its outputs. It will use the mappings in `cdb.addl_info["cui2<ont>"]`
     are present.
 
+    If set to "auto" (or missign), the value will be inferred from available
```
alhendrickson, line 262: Typo missign

medcat-v2/medcat/cat.py
```diff
@@ -532,8 +532,11 @@ def _get_entity(self, ent: MutableEntity,
         # addons:
         out_dict.update(self.get_addon_output(ent))  # type: ignore
         # other ontologies
-        if self.config.general.map_to_other_ontologies:
-            for ont in self.config.general.map_to_other_ontologies:
+        other_onts = self.config.general.map_to_other_ontologies
```
alhendrickson, line 535: Minor, I'm kind of feeling like as you already have the if statement in the function, you can just always call it to keep it simple here.


```python
other_onts =self._set_and_get_mapped_ontologies() 
for ont in other_onts
...

# Then in the function

  other_onts = self.config.general.map_to_other_ontologies
        if other_onts == "auto": # It always does this check anyway
```

As the function you wrote basically does the same if statement, and after the first run it isn't auto anymore.

As the alternative, remove the if statement from the function and trust that it's only called at the right time.

medcat-v2/medcat/cat.py
```diff
@@ -553,6 +556,28 @@ def _get_entity(self, ent: MutableEntity,
                 out_dict[ont] = ont_values  # type: ignore
         return out_dict
 
+    def _set_and_get_mapped_ontologies(
+            self,
+            ignore_list: list[str] = ["ontologies", "original_names",
```
alhendrickson, line 561: Minor - probably can use set[str] for the ignore list?

medcat-v2/tests/test_cat.py
```diff
@@ -133,6 +133,41 @@ def test_can_merge_config(self):
...
+    MY_ONT_MAPPING = {
+        # mapping doens't matter here, really
+        "ABC": "BBC"
+    }
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # add "mapping"
+        cls.model.cdb.addl_info[f"cui2{cls.MY_ONT_NAME}"] = cls.MY_ONT_MAPPING
```
alhendrickson, line 157: Minor - can add an explicit test test for the ignore_list and ignore_empty parameters for completeness

alhendrickson (approved):
lgtm

medcat-v2/medcat/config/config.py
mart-r: No, the typo is not missing!

PR #165, opened 2025-10-08
medact(feat): cdb utils for merging and navigation using pt2ch relations

tomolopolis:
- merge_cdb: migrated from v1, refactored and tested for v2
- other utils: ported from trainer / `medcat_utils.py`, documented and tested , as they are to be used elsewhere


mart-r:
If you resync with main, this should pass workflow since there's no need to test on 3.9 anymore.

PR #171, opened 2025-10-14
build(medcat): CU-869aujr7h Add nightly workflow to check library stability

mart-r:
Currently, most of our workflows run based on the dependencies defined in `uv.lock`. This is great for reproducibility, stabiliy, as well as speed of installation. However, it means that we're not checking the library against any of the new versions of our dependencies that come up over time. What this means is that even if our workflows run successfully, we don't have any confidence that a user installing the project will be able to actually use the project with the installed dependencies.

So this PR introduces a new - nightly - workflow that ignores the lock file and installs the latest available dependencies and runs some typical tests against those. The idea is to use this to catch any dependencies that have updates and become uncomatible for any reason.

As an example, `transformers==4.57.0` was released on 3rd of October. And it - inadvertently - dropped support for python 3.9. And because our workflows were not testing against this version, they were all passing. Yet a user installing after that date would have installed the version of `transformers` and found them to be incompatible only at run time.

The idea is that as a result of this PR failing, we can take action to identify the incompatible dependency and either patch the compatibility issue or release a patch version with an explicitly disallowed incompatible version.

The reason we don't _always_ (i.e in all worklows) test against the latest dependencies is so that we can separate the compatibility of a change from a PR (what existing workflows handle) from the stability of the overall library (what this new workflow handled). After all, we don't want to force a contributer to make drastic changes to dependencies and / or patch incompatibility issues unrelated to their change just to make the workflows run.

What the workflow does overall:
- Runs every night at 3am
- Runs over all our supported python versions (~~3.9~~, 3.10, 3.11, 3.12, 3.13)
- Runs over a number of different OS targets (ubuntu, macOs, Windows)
- Runs linting and typing
- Runs tests
- Runs regression test suite

There is a small caveat for MacOS runner:
- It ignores some DeID tests
  - The MacOS runner only has 7GB of RAM, and these would normally tip it over the edge
  - So currently they're just ignored on that (and only that) platform and only in CI
  - There's an extra 8 tests (i.e parts of a `TestCase`) that are ignored on MacOS

What it omits from the regular workflow is the backwards compatibility check. Though we may want to include that as well. I just felt like not doing all of it since this is going to be running on 12 different runners already (4 versions and 3 OSs).

### DIFF
--- a/medcat-v2/tests/utils/ner/test_deid.py
+++ b/medcat-v2/tests/utils/ner/test_deid.py
@@ -36,7 +36,7 @@ cnf = Config()
 cnf.general.nlp.provider = 'spacy'
 
 
-def should_do_test_ci() -> bool:
+def is_macos_on_ci() -> bool:
     return os.getenv("RUNNER_OS", "None").lower() != "macos"
 
 
@@ -116,7 +116,7 @@ def _train_model_once() -> tuple[tuple[Any, Any, Any], deid.DeIdModel]:
     return retval, model
 
 
-if should_do_test_ci():
+if is_macos_on_ci():
     _TRAINED_MODEL_AND_INFO = _train_model_once()
 
 
@@ -124,7 +124,7 @@ def train_model_once() -> tuple[tuple[Any, Any, Any], deid.DeIdModel]:
     return _TRAINED_MODEL_AND_INFO
 
 
-@unittest.skipIf(not should_do_test_ci(),
+@unittest.skipIf(not is_macos_on_ci(),
                  "MacOS on workflow doesn't have enough memory")
 class DeIDModelTests(unittest.TestCase):
     save_folder = os.path.join("results", "final_model")
@@ -178,7 +178,7 @@ Seen by Dr. M. Sully on 11/11/1996.
 '''  # noqa
 
 
-@unittest.skipIf(not should_do_test_ci(),
+@unittest.skipIf(not is_macos_on_ci(),
                  "MacOS on workflow doesn't have enough memory")
 class DeIDModelWorks(unittest.TestCase):
     save_folder = os.path.join("results", "final_model")
--- a/medcat-v2/tests/components/ner/trf/test_transformers_ner.py
+++ b/medcat-v2/tests/components/ner/trf/test_transformers_ner.py
@@ -20,7 +20,7 @@ import unittest.mock
 from ...addons.meta_cat.test_meta_cat import FakeTokenizer
 from ....pipeline.test_pipeline import FakeCDB, Config
 from .... import RESOURCES_PATH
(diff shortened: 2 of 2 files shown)


.github/workflows/medcat-v2-lib-stabiliy.yml
```diff
@@ -0,0 +1,58 @@
...
+    working-directory: ./medcat-v2
+
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
alhendrickson: From a design point of view:

I'm thinking in general you can make this as close to the _main one as possible? Keeping it simple and easy to maintain/

(Great to try all the other OS's just in this not main though - so this is one diff that is good)


But taking this example for the Test step, can we make them completely identical? Either change main or this one - as I dont want to have to think about if timeout-minutes is better than timeout (as an example of keeping it simple)

In here:
```
      - name: Test
        run: |
          uv run --python ${{ matrix.python-version }} python -m unittest discover
        timeout-minutes: 45
```

Vs in main
```
      - name: Test
        run: |
          timeout 30m uv run python -m unittest discover
```

medcat-v2/tests/utils/ner/test_deid.py
```diff
@@ -112,13 +116,16 @@ def _train_model_once() -> tuple[tuple[Any, Any, Any], deid.DeIdModel]:
...
+if should_do_test_ci():
+    _TRAINED_MODEL_AND_INFO = _train_model_once()
 
 
 def train_model_once() -> tuple[tuple[Any, Any, Any], deid.DeIdModel]:
     return _TRAINED_MODEL_AND_INFO
 
 
+@unittest.skipIf(not should_do_test_ci(),
+                 "MacOS on workflow doesn't have enough memory")
```
alhendrickson: So minor - could you alternatively find if there is enough available memory to a system? As I'm guessing this change will stop the test running on your mac locally as well

alhendrickson (approved):
Looks great, would be good to see what happens at 3am tonight

.github/workflows/medcat-v2-lib-stabiliy.yml
mart-r: The only reason the specific example was changed was because that didn't work for the Windows runner since it didn't have the `timeout` command available. And I didn't think it was within the scope here to change the main workflow to be in line with this change.

From a brief look we should be able to fully modularise this. I.e have the normal workflow steps (linting, typing, tests) in a separate workflow file (that never runs on its own) and we use it for both the main workflow and this one, but with slightly different inputs. I think it's probably worth doing. But I'd leave it for another task.

PR #177, opened 2025-10-16
bug(medcat): CU-869avau57 Fix cui to original names

mart-r:
The `cui2original_names` map was (in v1) in `cdb.addl_info`. But in v2 it's designed to reside in each CUI's `CUIInfo`. And when creating a new model from scratch it does. However, previsouly the legacy conversion wouldn't migrate the data correctly.

So this PR addresses that by:
- Fixing the legacy converter and making it move the data to `cdb.cui2info` for each CUI
- Adding an on the fly fix for CDBs that load with the incorrect conversion

The PR also adds a few tests that make sure that all of the above works as expected.


tomolopolis (approved):
lgtm

PR #178, opened 2025-10-16
build(medcat-den): Fix duplicate push versions to TestPyPI

mart-r:
Set pretend version with timesamp before build for den.

PR #174 was supposed to fix that, but the setting of the pretend version was after building so it didn't take effect at the right time.

This PR should fix that.


tomolopolis (approved):
lgtm

PR #181, opened 2025-10-20
build(medcat): CU-869awf45h Release after pre release

mart-r:
This will allow a minor release after a pre-release to be pushed out without any cherry-picked cahnges. Normally, with a patch release, you would expect to cherry-pick at least something, but with a pre-release that is being spun out as a full release, that's not always the case.

So this PR will allow no cherry-picks only if the version being created is a minor release after a pre-release for said minor release.

PS: The naming of the 2 release scripts isn't great. The prepare_patch_release.sh script is designed to deal with anything that already had a release within the same minor release cycle (i.e 2.2.1 after 2.2.0, but also 2.2.0 after 2.2.0a1 or 2.2.0rc1).

### DIFF
--- a/medcat-v2/.release/prepare_patch_release.sh
+++ b/medcat-v2/.release/prepare_patch_release.sh
@@ -117,8 +117,22 @@ if $DRY_RUN; then
     git checkout "$ORIGINAL_REF" >/dev/null 2>&1 || error_exit "Could not return to original branch"
 fi
 
+# Validate cherry-pick hashes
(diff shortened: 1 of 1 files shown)


tomolopolis:
Task linked: [CU-869awf45h Fix script for release without changes after a pre-release](https://app.clickup.com/t/869awf45h)

mart-r:
Accidentally had some other commits in there as well. Will close in favour of #182

PR #184, opened 2025-10-20
CU-869ag3k8a: Avoid using protected attribute for pipe getting

mart-r:
Small PR to avoid using protected (`_`-prefixed) attributes for getting pipe (or other activities).


tomolopolis:
Task linked: [CU-869ag3k8a Avoid using protected attributes in tutorials](https://app.clickup.com/t/869ag3k8a)

PR #188, opened 2025-10-20
CU-869awpaf3: (chore): medcat-trainer: remove cdb_utils, fix vestigia…

tomolopolis:
…l medcat-v2 migration issues


mart-r (approved):
Looks good

PR #201, opened 2025-10-28
feat(medcat-service): Add Gunicorn CLI args for max requests, jitter, and any other input from env

alhendrickson:
## New behaviour
- Support the two gunicorn args from this PR Ported from https://github.com/CogStack/cogstack-nlp/pull/141
- Support any other gunicorn args with an env var
- Print the whole command Gunicorn will run with to console

### Usage

Given this env var in the yaml
```
    environment:
      - SERVER_GUNICORN_EXTRA_ARGS=--worker-connections 1 --backlog 5
```

See this log line
```
cogstack-medcat-service-production  | Starting up the service using gunicorn server ...
cogstack-medcat-service-production  | + exec gunicorn --bind 0.0.0.0:5000 --workers=1 --threads=1 --timeout=300 '--access-logformat=%(t)s [ACCESS] %(h)s "%(r)s" %(s)s "%(f)s" "%(a)s"' --access-logfile=- --error-logfile=- --log-level info --config /cat/config.py --max-requests=1000 --max-requests-jitter=50 --worker-connections 1 --backlog 5 --worker-class uvicorn.workers.UvicornWorker medcat_service.main:app
cogstack-medcat-service-production  | [2025-10-29 09:22:43 +0000] [1] [INFO] Starting gunicorn 23.0.0
```

### DIFF
--- a/medcat-service/start_service_production.sh
+++ b/medcat-service/start_service_production.sh
@@ -35,12 +35,12 @@ fi
 
 if [ -z ${SERVER_GUNICORN_MAX_REQUESTS+x} ]; then
   SERVER_GUNICORN_MAX_REQUESTS=1000;
-  echo "SERVER_GUNICORN_MAX_REQUESTS is unset -- setting to default (sec): $SERVER_GUNICORN_MAX_REQUESTS";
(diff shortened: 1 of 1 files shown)


mart-r (approved):
Looks good to me!
Getting this in without #141 makes everything else simpler!

PR #204, opened 2025-10-29
CU-869azvxkn: Remove duplicate lines for transformers dependency.

mart-r:
Looks like both #167 and #171 added these lines and they now appeared duplicated in pyproject.toml. This PR removes the duplicates.

### DIFF
--- a/medcat-v2/pyproject.toml
+++ b/medcat-v2/pyproject.toml
@@ -89,8 +89,6 @@ spacy = [
 ]
 meta_cat = [
   "transformers>=4.41.0,<5.0", # avoid major bump
-  # Transformers 4.57 doesn't support 3.9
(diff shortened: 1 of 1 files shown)


medcat-v2/pyproject.toml
```diff
@@ -103,8 +103,6 @@ dict_ner = [
 ]
 deid = [
   "datasets>=2.2.2,<3.0.0",
-  # Transformers 4.57 doesn't support 3.9
-  "transformers!=4.57.0; python_version == '3.9'",
```
alhendrickson: As a loose thought, could you even remove the non-duplicated line now? If we don't specifically support 3.9 anymore. May as well keep it I guess just a thought.

PR #207, opened 2025-10-29
feat(medcat-service): Add optional metrics instrumentation with prometheus

alhendrickson:
Add prometheus metrics for collection on a `/metrics` API. Reasoning is I want to know requests per second, and average request duration across N instances of the service

## Usage
Toggle on or off without the env variable

```
2025-10-29 13:42:15,144 [ACCESS] - 127.0.0.1:51960 - "GET /metrics HTTP/1.1" 404 Not Found
```

```
export APP_ENABLE_PROMETHEUS=true

2025-10-29 13:43:16,278 [ACCESS] - 127.0.0.1:54424 - "GET /metrics HTTP/1.1" 200 OK
```

Metrics can then be collected and shown in grafana, for example: 

image


mart-r (approved):
Yeah, a lot of this is just linting. But overall straightforward.

PR #217, opened 2025-11-06
bug(medcat): CU-869b36xv7 Avoid meta cat issue when getting model card

mart-r:
When getting the model card, MetaCAT is imported in order to include it in the model card (if it exists). However, if it doesn'st exist an exception is raised since the import fails due to an underlying `ImportError`.

This PR allows getting the model card without the optional extra.


tomolopolis:
Task linked: [CU-869b36xv7 Avoid issue with non-existent MetaCAT wheng getting model card](https://app.clickup.com/t/869b36xv7)

PR #223, opened 2025-11-12
fix/feat(frontend): Fix container appearing outside of viewport & improvement on responsive layouts

dcstang:
## Summary 
This is a bugfix that lead into minor improvements on responsive layouts for the AnonCAT demo app.

**Problem**: the output text container was appearing outside of viewport for certain screen sizes

**Fix**: Retaining mostly the default layout, with UI/UX Changes for tablet and mobile viewports.

## Changes

- Vue grid system syntax fix & CSS fix for overflow and displaying containers outside of viewport
- Set up responsive design with media queries
- Reordered buttons for better cohesion with cogstack color theme on mobile (Deidentify and Use Example buttons first, then Redact checkbox below)
- Reduced padding and gaps throughout mobile layout for more compact design
- Fixed scrolling issues on viewports <1280px by properly constraining viewport-container height
- Reduced border sizes on mobile from 15px to 5px for cleaner appearance
- Reorganised mobile middle container for logo to be a watermark background

Additional bug fix
- Removed unused import that called defunct variable (deid_text) from views.py

### Before (Desktop)
before-anoncat-desktop
The output text container is **displayed outside of viewport** and not below fold

### After (Desktop)
image

### After (Mobile)
image

## Test plan
- [x] Tested mobile view (<768px)
- [x] Tested tablet view (768-1023px)
- [x] Verified scrolling works correctly

## Additional Notes
All changes are CSS-only for the responsive layouts. No functional changes to the deidentification logic.
Discussion whether to update this to medcat v2 as well? currently on 1.1
At some point to update the currently deployed version that is on internal infra.

PR #227, opened 2025-11-18
CU-869b75fhg: Fix issue with actual DeID in anon cat demo.

mart-r:
The deid_text method in medcat.utils.ner was deprecated in favour of using the  instead. However, the change in #223 removed the import (probably because otherwise it failed due to an import error) without replacing it with the new alteranative. This PR fixes that.


tomolopolis:
Task linked: [CU-869b75fhg Fix anoncat demo app](https://app.clickup.com/t/869b75fhg)

PR #386, opened 2026-03-31
deps(medcat+medcat-service): CU-869cqbd24 Allow transformers 5

mart-r:
This PR allows transformers 5 for MedCAT and medcat-service. Let's see if it works.

Again, posted due to the dependabot hounding on this.

EDIT:
The shorthand imports should already work in `transformers==4.x` as well.


mart-r:
Won't be workign on this for hte time being

PR #400, opened 2026-04-07
perf: fix low-hanging performance issues in MetaCAT and linking

bgriffen:
## Summary

Four small, safe performance fixes across the MetaCAT and linking components:

- **Scope batch padding to current batch** (`ml_utils.py`) — `create_batch_piped_data` was computing `max_seq_len` over the entire dataset on every batch call, then slicing `data[start_ind:end_ind]` three times. Now scoped to a single batch slice. Reduces unnecessary padding and iteration.

- **In-place similarity update** (`vector_context_model.py`) — `_preprocess_disamb_similarities` was copying the similarities list, clearing it, and rebuilding via list comprehension. Replaced with a simple in-place loop, eliminating three intermediate allocations in the disambiguation hot path.

- **O(1) dict lookup instead of O(n) values scan** (`data_utils.py`) — `undersample_data` and `encode_category_values` checked membership against `category_value2id.values()` (linear scan) every iteration. Since `label_data` dicts use the same keys, check against the dict directly.

- **Append instead of list concatenation** (`ml_utils.py`) — `_eval_predictions` used `dict.get(k, []) + [item]` which allocates a new list per iteration. Switched to `setdefault` + `append` for O(1) amortized insertions.

All changes are behaviour-preserving — no new dependencies, no config changes, no API changes.

## Test plan

- [ ] Existing `medcat-v2` unit tests pass across Python 3.10–3.13
- [ ] MetaCAT training + inference produces identical results
- [ ] Linking disambiguation produces identical concept rankings

### DIFF
--- a/medcat-v2/medcat/components/addons/meta_cat/data_utils.py
+++ b/medcat-v2/medcat/components/addons/meta_cat/data_utils.py
@@ -315,10 +315,9 @@ def undersample_data(data: list, category_value2id: dict, label_data_,
             label_data_counter[sample[-1]] += 1
 
     label_data = {v: 0 for v in category_value2id.values()}
-    for i in range(len(data_undersampled)):
-        if data_undersampled[i][2] in category_value2id.values():
-            label_data[data_undersampled[i][2]] = (
-                label_data[data_undersampled[i][2]] + 1)
+    for sample in data_undersampled:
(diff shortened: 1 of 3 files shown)


mart-r (approved):
Thanks for the contribution!

Looks good to go. Minor changes, but clear improvement.

PR #402, opened 2026-04-07
bug(medcat-trainer): CU-869ctq789 Fix saving model packs

mart-r:
Hopefully this fixes:

https://discourse.cogstack.org/t/unable-to-save-model-from-trainer/378

The idea is to allow model pack based projects to overwrite their model packs as the CDB based ones were designed to.
Though notably without the `overwrite=True` option they, too, would probably have failed.

So this PR does the following:
- Creates a separate function to do the model pack overwriting if/when required
- Moves the old model pack to a temporary folder
  - Because the `CAT.save_model_pack` method doesn't allow (or delegate the allowing of) overwriting existing ifles
  - And if we just removed the previous file(s) I'd be afraid of leaving a state with a model pack path that refers to nothing
  - If all goes well, the copy is destroyed after
  - If something fails, attempt to move the old one back to its original location

### DIFF
--- a/medcat-trainer/webapp/api/api/views.py
+++ b/medcat-trainer/webapp/api/api/views.py
@@ -1,8 +1,9 @@
 import logging
 import os
 from smtplib import SMTPException
-from tempfile import NamedTemporaryFile
+from tempfile import NamedTemporaryFile, TemporaryDirectory
 from typing import Any
+import shutil
 
 from background_task.models import Task, CompletedTask
 from django.contrib.auth.views import PasswordResetView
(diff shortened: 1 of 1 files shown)


medcat-trainer/webapp/api/api/views.py
```diff
@@ -579,11 +580,39 @@ def save_models(request):
...
+        # to the specific path rather than into the folder
+        temp_folder = os.path.join(tmp_dir, "model_copy")
+        shutil.move(model_path, temp_folder)
+        try:
+            cat.save_model_pack(
+                os.path.dirname(model_path),
+                pack_name=os.path.basename(model_path),
+                add_hash_to_pack_name=False)
+        except Exception as e:
+            logger.warning("Unable to save model pack. Restoring previous state")
```
alhendrickson: Can you log the exception as well here? As I think you'll never see why it failed if the restore original doesnt error.

medcat-trainer/webapp/api/api/views.py
```diff
@@ -579,11 +580,39 @@ def save_models(request):
     project = ProjectAnnotateEntities.objects.get(id=p_id)
     cat = get_medcat(project=project)
 
-    cat.cdb.save(project.concept_db.cdb_file.path)
+    if project.concept_db is not None:
+        # CDB / vocab based
+        cat.cdb.save(project.concept_db.cdb_file.path, overwrite=True)
+    else:
+        _overwrite_model_pack(cat, project.model_pack.path)
```
alhendrickson, line 588: Minor - I think addding  the matching comment `# Model Pack Based`  would help clarify it here instead of it being implied.

Extremely minor, I feel like ideally there would be some enum in project like `project.type == "MODEL_PACK` or whatever as I think I've had to do similar checks

alhendrickson (approved):
Looks good to me

medcat-trainer/webapp/api/api/views.py
```diff
@@ -579,11 +580,39 @@ def save_models(request):
...
 
     return Response({'message': 'Models saved'})
 
 
+def _overwrite_model_pack(cat, model_path: str):
+    # NOTE: cannot overwrite, so working around
```
alhendrickson, line 594: Minor - maybe gen a doc string explaining why you cant overwrite the model pack

medcat-trainer/webapp/api/api/views.py
mart-r: I think we may ditch the CDB / Vocab based version altogether. But regardless, a problem for another day.

alhendrickson (approved):
Cheers - lgtm

PR #408, opened 2026-04-09
bug(medcat): CU-869cunfx7 Fix supervised training order of operations issue

mart-r:
The previous PR #374 made a change where the call to the `Pipeline.pipeline.entity_from_tokens` method (previously called inside the `Trainer.add_and_train_concept` method) outside the try/except block in the `for` loop in the `Trainer._train_supervised_for_project2` method. So while previously issues where no tokens were found for a specific span were logged as warnings (at least normally), the change meant that these were raised instead.

This could happen if an annotated dataset had an annotation mid-token. For example, if trying to get the entity for `rb` in `orbits` the process will fail because the entire word is one token and the subpart of it cannot be accurately represented by any tokens.

So what this PR does:
- Fixes some minor deprecation warning issues from #374
  - Deprecation warnings were added to various tokenizer `entity_from_tokens` methods
    - Even though these remain to be used if/when there is no pre-existing entity corresponding to the span
    - So therefore these were now removed
  - Deprecation warning was NOT added to the `Pipeline.entity_from_tokens` method
    - This is what was actually not supposed to be used directly
    - So this deprecation warning was added
- Fixed the change by catching the exception and logging a warning like before
  - This involved separating out the warning logic so it can be used in different places
  - Though an exception can still be raised if `Trainer.strict_train = True` is set
- Adds a few tests to verfiy behaviour
  - Make sure that the exception is normally (`Trainer.strict_train == False`) not raised
  - And make sure that the exception is raised when `Trainer.strict_train` is true
  

The idea of the PR is that the first workflow will fail (it adds the tests, but not the actual fix).
And the subsequent one will succeed.

### DIFF
--- a/medcat-v2/medcat/pipeline/pipeline.py
+++ b/medcat-v2/medcat/pipeline/pipeline.py
@@ -345,6 +345,14 @@ class Pipeline:
         Returns:
             MutableEntity: The resulting entity.
         """
+        warnings.warn(
+            "The `medcat.pipeline.pipeline.Pipeline.entity_from_tokens` method is"
+            "depreacated is subject to removal in a future release. Please use "
+            "`medcat.pipeline.pipeline.Pipeline.entity_from_tokens_in_doc` "
+            "instead.",
+            DeprecationWarning,
+            stacklevel=2
+        )
         return self._tokenizer.entity_from_tokens(tokens)
 
     def entity_from_tokens_in_doc(self, tokens: list[MutableToken],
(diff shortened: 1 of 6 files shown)


adam-sutton-1992 (approved):
LGTM.

I'm a bit unsure why the depreciation warnings were removed where they were, or why it was still called on the in replace method in the first place.

If you think this is more future / error proof then I'm cool with it.

mart-r:
> I'm a bit unsure why the depreciation warnings were removed where they were, or why it was still called on the in replace method in the first place.

As per the PR description:

> - Fixes some minor deprecation warning issues from #374
>  - Deprecation warnings were added to various tokenizer `entity_from_tokens` methods
>    - Even though these remain to be used if/when there is no pre-existing entity corresponding to the span
>    - So therefore these were now removed
>  - Deprecation warning was NOT added to the `Pipeline.entity_from_tokens` method
>    - This is what was actually not supposed to be used directly
>    - So this deprecation warning was added

In other words the methods that had the deprecation warning were still being legitimately used. They shouldn't be called directly instead of the other ones, but they are (and will continue to be) still a part of the API.

PR #414, opened 2026-04-21
feat(medcat):CU-869cy3xa0 Improve training

mart-r:
This PR does an overhaul to the training setup of MedCAT:
- It modifies the existing `TrainableComponent` protocol to also include a `train_unsupervised` method
  - And uses that over the "check config for train and run inference" unsupervised training
- It allows all components that follow the `TrainableComponent` protocol to be trained supervised
  - Previously only the linker was able to be trained in a supervised manner
- It provides a few utility methods to allow training and evaluating components individually
  - I.e dataset-aware components that will enable either training or evaluating only NER or Linker if/when required

Example code snippets:
1. When only training linker
```python
with dataset_aware_component(cat, CoreComponentType.ner, DATASET):
    trainer.train_supervised_raw(DATASET, nepochs=1)
```
2. When only training NER
```python
with dataset_aware_component(cat, CoreComponentType.linking, self.DATASET):
    trainer.train_unsupervised([doc['text'] for proj in self.DATASET['projects'] for doc in proj['documents']], nepochs=1)
```
3. When doing evaluation / stats one component at a time
```python
with dataset_aware_component(cat, CoreComponentType.ner, self.DATASET):
    tps, fns, tps, cui_prec, cui_rec, cui_f1, cui_counts, examples = get_stats(
        cat, self.DATASET, do_print=False)
```

### DIFF
--- a/medcat-v2/medcat/trainer.py
+++ b/medcat-v2/medcat/trainer.py
@@ -64,10 +64,8 @@ class Trainer:
         with self.config.meta.prepare_and_report_training(
             data_iterator, nepochs, False
         ) as wrapped_iter:
-            with temp_changed_config(self.config.components.linking,
-                                     'train', True):
-                self._train_unsupervised(wrapped_iter, nepochs, fine_tune,
-                                         progress_print)
+            self._train_unsupervised(wrapped_iter, nepochs, fine_tune,
+                                     progress_print)
 
     def _train_unsupervised(self,
                             data_iterator: Iterable,
@@ -91,11 +89,18 @@ class Trainer:
                 # Convert to string
                 line = str(line).strip()
 
+
+                # inference run for the document
                 try:
-                    _ = self.caller(line)
+                    doc = self.caller(line)
                 except Exception as e:
                     logger.warning("LINE: '%s...' \t WAS SKIPPED", line[0:100])
                     logger.warning("BECAUSE OF:", exc_info=e)
+                    continue
+                for comp in self._pipeline.iter_all_components():
+                    if isinstance(comp, TrainableComponent):
+                        logger.debug("Training on component %s", comp.full_name)
+                        comp.train_unsupervised(doc)
             else:
                 logger.warning("EMPTY LINE WAS DETECTED AND SKIPPED")
 
--- a/medcat-v2/medcat/trainer.py
+++ b/medcat-v2/medcat/trainer.py
@@ -15,7 +15,7 @@ from medcat.data.mctexport import (
     MedCATTrainerExport, MedCATTrainerExportAnnotation, MedCATTrainerExportProject,
     MedCATTrainerExportDocument, count_all_annotations, iter_anns)
 from medcat.preprocessors.cleaners import prepare_name, NameDescriptor
-from medcat.components.types import CoreComponentType, TrainableComponent
+from medcat.components.types import TrainableComponent
 from medcat.components.addons.addons import AddonComponent
 from medcat.pipeline import Pipeline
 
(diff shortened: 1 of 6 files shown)


medcat-v2/medcat/trainer.py
```diff
@@ -637,18 +637,15 @@ def add_and_train_concept(self,
...
+        for component in self._pipeline.iter_all_components():
+            if not isinstance(component, TrainableComponent):
+                continue
             # Train Linking
             if isinstance(mut_entity, list):
                 mut_entity = self._pipeline.entity_from_tokens(mut_entity)
-            linker.train(cui=cui, entity=mut_entity, doc=mut_doc,
-                         negative=negative, names=names)
+            component.train(cui=cui, entity=mut_entity, doc=mut_doc,
+                            negative=negative, names=names)
```
adam-sutton-1992, line 648: Does this mean we're still doing unsupervised on a per entity basis? I can't think of a case where in an unsupervised manner you would need the entity.

adam-sutton-1992:
A few queries but I think it looks good. I might've missed these within the commits:

If you have two trainable components. is it possible to turn of training for one of them when running training methods? Do the dataset aware components serve that purpose?


And one more above^^^

medcat-v2/medcat/trainer.py
mart-r: This is supervised training still. This is within the `add_and_train_concept` method.

mart-r:
> If you have two trainable components. is it possible to turn of training for one of them when running training methods? Do the dataset aware components serve that purpose?

The description already had 2 examples for this :) 

The dataset aware implementation can serve that purpose. Because they replace the specific component with another one (which isn't trainable, but that's kind of irrelevant since it's a different component) for the duration of the context manager.

But I think what makes it unclear is that in the example I've given it a dataset, but realistically, you could provide an empty dataset for it, i.e like this:

```python
# supervised
with dataset_aware_component(cat, CoreComponentType.ner, {"projects" : []}):
    trainer.train_supervised_raw(DATASET, nepochs=1)
# unsupervised
with dataset_aware_component(cat, CoreComponentType.ner, {"projects" : []}):
    trainer.train_unsupervisedsupervised(["list", "of", "texts'], nepochs=1)
```

medcat-v2/tests/utils/test_training_utils.py
```diff
@@ -238,3 +250,29 @@ def test_train_unsupervised_can_train_only_ner_when_linker_is_cheating(self):
...
+        linker = _TrainablePassThroughLinker()
+        cat = _FakeCat(self.DATASET, [ner, linker])
+        trainer = Trainer(cat.cdb, cat.__call__, cat.pipe)
+
+        with unittest.mock.patch("medcat.trainer.prepare_name", return_value={"abc": {}}):
+            with dataset_aware_component(cat, CoreComponentType.linking, self.DATASET):
+                trainer.train_supervised_raw(self.DATASET, disable_progress=True)
+
+        self.assertEqual(ner.sup_train_calls, 1)
+        self.assertEqual(linker.sup_train_calls, 0)
```
adam-sutton-1992, line 278: Why not train both at the same time?

adam-sutton-1992 (commented):
Just the one comment:

medcat-v2/tests/utils/test_training_utils.py
mart-r: You _can_. But the point is that you don't have to! I.e flexibiliy.

adam-sutton-1992: Sorry I missread this. assume you intended for only one or the other.

PR #416, opened 2026-04-23
fix(embedding-linker) optimizer, transformer config, and naming fixes

adam-sutton-1992:
Hihi,

Three fixes/changes in here for the most part.

1. The embedding linker optimzer doesn't reset upon providing new models. Training would be difficult. So I expose a "load_transformers" to both static and trainable linkers. The trainable model does a bit more with `reset_optimizer_and_scaler` which points the optimzer to the right model. I also expose learning rate and weight decay, so they're config settings that can be set and changed here.
2. The naming of the embedding linker was inconsistent. So I've set `name` to `trainable_embedding_linker` and the class name to Linker. To be consistent with previous design. This was an issue with creating new components and deserializing.
3. With embed_cuis and embed_names being tucked away in the context_model, it didn't make sense to have the model being set there. Somewhat missed from just its static linking days.


medcat-plugins/embedding-linker/src/medcat_embedding_linker/trainable_embedding_linker.py
```diff
@@ -366,10 +395,14 @@ def create_new_component(
         cdb: CDB,
         vocab: Vocab,
         model_load_path: Optional[str],
-    ) -> "TrainableEmbeddingLinker":
+    ) -> "Linker":
         return cls(cdb, cdb.config)
 
     def serialise_to(self, folder_path: str) -> None:
+        # Ensure final partial batch is not dropped before saving model state.
+        logger.info("Flushing final training batch before saving model.")
```
mart-r: It doesn't seem to be doing that here?

mart-r (approved):
The logged comment looks odd in `serialise_to` - what it's describing doesn't seem to be happening.

Other than that, looks good.

medcat-plugins/embedding-linker/src/medcat_embedding_linker/trainable_embedding_linker.py
adam-sutton-1992: Fixed

adam-sutton-1992:
Fixed that doc issue - from the old embedding linker.

PR #418, opened 2026-04-24
feat(medcat-den): Make credentials optional for remote dens

alhendrickson:
## Current Behavior
- I have a medcattery instance with no authentication, this is the default when I start it up.
- I can't use medcat-den without passing an unused credentials block

EG I have to do this in the json:

```
        "medcattery_prod": {
            "type": "medcattery",
            "host": "https://medcattery.example.com",
            "credentials": {
                "unused: "unused"
            }
        }
```

## New behavior

I should now be able to skip that credentials block altogether
```
        "medcattery_prod": {
            "type": "medcattery",
            "host": "https://medcattery.example.com"
}

```

(Note I also need to change medcat-den-medcattery implementation)

### DIFF
--- a/medcat-den/src/medcat_den/config.py
+++ b/medcat-den/src/medcat_den/config.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Union, Optional
 from pathlib import Path
 
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


mart-r (approved):
Yeah, why not. I've had the same issue before... So this would certain simplify some stuff

PR #419, opened 2026-04-24
Fix issue with PR paths for embedding linker workflow

mart-r:
The previous path was incorrect which meant the workflow didn't run at PR time.

### DIFF
--- a/.github/workflows/medcat-embedding-linker_ci.yml
+++ b/.github/workflows/medcat-embedding-linker_ci.yml
@@ -7,7 +7,7 @@ on:
       - 'medcat-embedding-linker/v*.*.*'  
   pull_request:
     paths:
-      - 'medcat-plugins/medcat-embedding-linker/**'
+      - 'medcat-plugins/embedding-linker/**'
       - '.github/workflows/medcat-embedding-linker**'
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/medcat-embedding-linker_ci.yml
```diff
@@ -8,7 +8,7 @@ on:
   pull_request:
     paths:
       - 'medcat-plugins/medcat-embedding-linker/**'
-      - '.github/workflows/medcat-embedding-linker**'
+      - '.github/workflows/embedding-linker**'
```
alhendrickson: Hey I dont think this is right?

The file name is `.github/workflows/medcat-embedding-linker_ci.yml` ?

alhendrickson (changes requested):
Is this right?

.github/workflows/medcat-embedding-linker_ci.yml
mart-r: Yep - you're right - fixed the wrong one

mart-r:
PS: 
The path is correct now. We can see that from the fact that the workflow runs and fails due to the unfixed issue of #420.

alhendrickson (approved):
Approved - feels reasonable to merge this in, and then prove the other branch makes the build pass again

PR #420, opened 2026-04-24
fix(embedding_linker)fixing the comp name requirement

adam-sutton-1992:
Yeah. MIssed this last PR.


mart-r:
Ideally, we'd wait for #419 to be correct and merged, then merge it in here and look at workflow to make sure there's no further tiny issues out there.

mart-r (approved):
Looks great!

PR #421, opened 2026-04-26
fix(embeddinglinker) - Embedding Mention Mask Fix and Embed Cui via longest name.

adam-sutton-1992:
Two fixes here leading to a performance upgrade.

1. Provide mention masks for embed_cui and name methods. I previously (and erroneously) thought that mean pooling would be the correct attention for names and cui embeddings, as that's the only thing there. However it does contain the <CLS> and <SEP> special tokens as part of it. Throwing off scoring - if you detected say "glucose", embed with no context and just scored that glucose vs an emebdded "glucose" the scoring could be quite far off. You would expect a similarity of 1, but this would return 0.875 here.
2. Embed CUIs via their longest name rather than their preferred name. Which solves issues with disambiguation. Another example to explain this; the name "glucose" is ambiguous. It could be a measurement (snomed: 33747003), or a substance (67079006). When embedding via preferred names; the preferred name for the measurement version is "Glucose measurement, blood", the subtance version is... "glucose".  The model will always choose substance over measurement due to cosine similarities. The longest names however makes it an easier task - improving performance.

These changes performance in metrics as such to:

Epoch: 0, Prec: 0.2862910530810839, Rec: 0.7398836198247609, F1: 0.4128382160850905.

An improvement of previous similar recalls which were around 0.66.

### DIFF
--- a/medcat-plugins/embedding-linker/src/medcat_embedding_linker/embedding_linker.py
+++ b/medcat-plugins/embedding-linker/src/medcat_embedding_linker/embedding_linker.py
@@ -585,7 +585,7 @@ class Linker(AbstractEntityProvidingComponent):
     def predict_entities(
         self, doc: MutableDocument, ents: list[MutableEntity] | None = None
     ) -> list[MutableEntity]:
-        if self.cnf_l.train and self.comp_name == "embedding_linker":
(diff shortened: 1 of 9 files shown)


adam-sutton-1992:
Closed as it included the transformer NER stuff. Which isn't yet ready.

PR #423, opened 2026-05-04
Mc trainer fix cui UI

vladd-bit:
Fixes:
 - cui/type id /synonyms displayed properly in the training annotator interface

I've added an extra start debug script but im not sure its necessary...

Also I didn't test anything else so I hope this doesn't break anything like annotating for meta-ants or relations .....


medcat-trainer/envs/env
```diff
@@ -2,7 +2,7 @@
 OPENBLAS_NUM_THREADS=1
 
 ### MedCAT cfg ###
-MEDCAT_CONFIG_FILE=/home/configs/base.txt
+MEDCAT_CONFIG_FILE=../../configs/base.txt # /home/configs/base.txt
```
mart-r: Is this supposed to just be a debug-time change?

medcat-trainer/envs/env
```diff
@@ -12,9 +12,13 @@ ENV=non-prod
 # Complete once this is deployed
 CSRF_TRUSTED_ORIGINS=
 
+USE_OIDC=0
+
 ### Django debug setting - to live-reload etc. ###
 DEBUG=1
 
+DJANGO_LOG_LEVEL=DEBUG
```
mart-r: Do we need to default to DEBUG logs in production? Wouldn't that just flood the logs?

medcat-trainer/docker-compose-dev.yml
```diff
@@ -84,5 +84,5 @@ volumes:
 
 networks:
   gateway-auth_gateway-net:
-    external: true
+  #  external: true
```
mart-r: Is there a reason to keep this but commented? Perhaps either remove or document why / when it should be enabled.

medcat-trainer/webapp/scripts/run_local_debug.sh
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bash
```
mart-r, line 1: This does a lot. And it makes a fair number of assumptions on the user's environment.

Perhaps we should include this in a `scripts` or `tools` folder away from the root of the project? And then have a README there to describe when this can be used (e.g prerequisites of `uv`, `docker`, specific python version), what it's for (i.e solo / local dev), and what its limitations are?

mart-r (changes requested):
I think overall this looks good!

But there's a few things that I think were debug-time changes that were committed.

And the local debug script should probably not be in the root of the project (and should include some more documentation as to when / why it can / should be used).

medcat-trainer/envs/env
vladd-bit: its the dev config env vars, prod not affected, but yes, ill revert

medcat-trainer/webapp/scripts/run_local_debug.sh
vladd-bit: sounds good, ill put it there

medcat-trainer/envs/env
vladd-bit: should be fine as its in dev env var, envs/env-prod is used in production

medcat-trainer/docker-compose-dev.yml
vladd-bit: ok ill doc this,  if left as it was it requires the network to exist already otherwise the dev compose wont start, I just assumed its not an essential component so... decided to comment it out

mart-r (approved):
Looks good to me.

PR #428, opened 2026-05-11
feat(medcat-trainer-client): Support OIDC in medcat trainer client

alhendrickson:


jocelyneholdbrook (approved):

medcat-trainer/client/mctclient.py
```diff
@@ -237,15 +288,31 @@ def __init__(self, server=None, username=None, password=None):
...
+            if keycloak_settings is None:
+                kc_settings = KeycloakSettings()
+            else:
+                if not isinstance(keycloak_settings, KeycloakSettings):
+                    raise TypeError("keycloak_settings must be a KeycloakSettings instance")
+                kc_settings = keycloak_settings
+
+            token = get_keycloak_access_token(kc_settings)
+            self.headers = {"Authorization": f"Bearer {token}"}
+            return
```
mart-r, line 305: I feel like the early return could bite us in the bum later down the road.
I.e I feel like the 2 different auth / header getting methods should be moved outside the `__init__` and called in if/else manner.

PR #430, opened 2026-05-11
fix(medcattrainer) - solve `irrelevant` annotations document submit error

silverfoxdoc:
In medcattrainer, if document annotations are marked as irrelevant, submitting a document throws a 500 error because MedCAT v2 uses a typed LinkingFilters object, not a dict.

```
TypeError: 'LinkingFilters' object does not support item assignment
cat.config.components.linking.filters['cuis_exclude'] = set()
```

Patch submitted as discussed on https://discourse.cogstack.org/t/medcat-trainer-issues-cant-submit-annotations-as-irrelevant/381


mart-r (approved):
Thank you for the contribution! It's much appreciated.

Looking good!

mart-r:
The workflow failure is fine given the context of an outside PR.

PR #434, opened 2026-05-12
build: Dependabot config to keep updates separated

alhendrickson:
This is an attempt to keep the github dependabot PRs separated per folder.

Unsure if this is the fix or if it's actually part of the security settings.


mart-r (approved):
Thanks!

Let's hope it does fix it...

PR #435, opened 2026-05-12
tests(medcat): CU-869d8wft4 Add tests to name count during supervised training

mart-r:
This seems to have not been working in MedCAT==2.5.3, see [this discourse thread](https://discourse.cogstack.org/t/new-synonyms-and-train-count-not-updating-general-query/384).

However, at a closer look it turned out this was in latest state of the codebase.
At a closer look it looks like this was fixed by #374 and released in v2.7.0 (April 1st).

I verified by checking that the issue does exist in v2.6.0 but does not in v2.7.0.

Nevertheless, I felt like adding tests for this would be appropriate.


adam-sutton-1992:
LGTM - you know the embedding linker doesn't increase these training counts, right?

adam-sutton-1992 (approved):

mart-r:
> LGTM - you know the embedding linker doesn't increase these training counts, right?

Yeah, that's fine. These are only really applicable for the vector context model.

Perhaps they should be kept separate from the core structure. But it is what it is for now.

PR #467, opened 2026-05-13
feat(medcat): CU-869d9n2rg Avoid running pipe twice for supervised training

mart-r:
This PR separates the callers used for supervised and self-supervised training.

The self-supervised training will run the entire pipe.
And the supervised training only runs the tokenizer.

Along the way also fixed up a few minor things
- The trainable component counting based warning (i.e when nothing to train)
- The entities sent to be trained supervised now have a detected name
  - Without it the name counts wouldn't be updated

<details>
<summary>Original(ish) description</summary>
If all goes well this just works out of the box.

EDIT:
Looks like it's not working since the linker unsupervised training needs the NER step to have run because otherwise it doesn't know what entities to train on. Need to figure out a solution for that.
</summary>

### DIFF
--- a/medcat-v2/medcat/trainer.py
+++ b/medcat-v2/medcat/trainer.py
@@ -28,11 +28,14 @@ logger = logging.getLogger(__name__)
 class Trainer:
     strict_train: bool = False
 
-    def __init__(self, cdb: CDB, caller: Callable[[str], MutableDocument],
+    def __init__(self, cdb: CDB,
+                 tokenizer: Callable[[str], MutableDocument],
+                 pipe: Callable[[str], MutableDocument],
                  pipeline: Pipeline):
         self.cdb = cdb
         self.config = cdb.config
-        self.caller = caller
+        self.tokenizer = tokenizer
+        self.pipe = pipe
         self._pipeline = pipeline
 
     def train_unsupervised(self,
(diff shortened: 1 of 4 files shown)


medcat-v2/medcat/trainer.py
```diff
@@ -419,7 +419,12 @@ def _prepare_doc_with_anns(
         for ann in anns:
             tkns = doc.get_tokens(ann['start'], ann['end'])
             try:
-                ents.append(self._pipeline.entity_from_tokens_in_doc(tkns, doc))
+                ent = self._pipeline.entity_from_tokens_in_doc(tkns, doc)
+                # avoid setting CUI so as to not show our hand
+                raw_name = ann['value'].lower().replace(
+                    " ", self.cdb.config.general.separator)
+                ent.detected_name = raw_name
+                ents.append(ent)
```
adam-sutton-1992, line 430: As in the correct CUI isn't already set when performing training?

adam-sutton-1992 (approved):
LGTM. Should be a big speedup.

One minor question below.

medcat-v2/medcat/trainer.py
mart-r: You're right - it doesn't matter. Even in the previous case this would have already gone through the pipe. Might as well add that as well.

PR #507, opened 2026-05-27
build(medcat-den and embedding-linker): CU-869ddh1jv Fix downstream installs

mart-r:
This PR fixes downstream installs related to #505 and #503.

That's because these depend on a release version of `medcat` that hadn't specified the incompatible `typer` versions.


alhendrickson (approved):
Approved 

Though if this bug has broken everything that uses medcat, maybe we need to add click into medcat directly instead

mart-r:
> Though if this bug has broken everything that uses medcat, maybe we need to add click into medcat directly instead

Sure, we could do that. But it would still require a release for it to be picked up.

PR #510, opened 2026-05-28
test(medcat-trainer): expand backend and frontend unit test coverage

tomolopolis:
Add Django API tests for utils, views, models, serializers, permissions, Solr/model cache, admin actions, and metrics. Add Vitest specs for core UI components, admin lists, filters, and runtime config. Refresh uv.lock for medcat 2.7.x alignment with pyproject constraints.


dcstang (approved):
looks good!

PR #515, opened 2026-05-28
fix(medcat-den): Fix wrapper usage of upstream trainer object

mart-r:
This PR fixes an issue with `medcat-den` introduced by `medcat==2.8.0`. The wrapper for the trainer needed to be slightly adjusted.

This should also fix the issue causing workflow to fail in #514 

EDIT:
@adam-sutton-1992 this may conflict with your specific changes you're working on. But I don't imagine it'd be a massive merge conflict.


tomolopolis (approved):
lgtm

PR #516, opened 2026-05-28
fix(medcat-embedding-linker): Fix issue causing workflows to fail

mart-r:
The tests require the trainable linker to follow the `TrainableComponent` protocol. But as of 2.8.0 the current implementation doesn't (since it doesn't declare the `train_unsupervised` method).

This PR fixes that.


tomolopolis (approved):
lgtm

PR #518, opened 2026-05-29
build(medcat and medcat-den and medcat-embedding-linker): Fix spacy-click issue

mart-r:
This PR will fix this once and for all by explicitly requiring `click`. Related to #505 and #507.

I didn't originally want to add `click` as a dependency since that wasn't the culprit. But `typer` already managed to get a subsequent release in, and as such broke all our installs again. So I'm going to bite the bullet and have it in here still.

EDIT:
Notably, the `typer` constraint isn't actually needed after this change. So may be better off removing it instead of widening it like I did now.

### DIFF
--- a/medcat-den/pyproject.toml
+++ b/medcat-den/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
     "diskcache>=5.6.3,<6.0",
     # NOTE: transitive dependency that breaks other deps
     #       e.g `spacy`'s dependency `click` isn't installed
-    "typer!=0.26.0,!=0.26.1,!=0.26.2",
+    "typer!=0.26.*",
     # transitive dependency that spacy needs but doesn't delcare...
     "click",
 ]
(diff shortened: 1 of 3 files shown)


medcat-den/pyproject.toml
```diff
@@ -35,7 +35,9 @@ dependencies = [
     "diskcache>=5.6.3,<6.0",
     # NOTE: transitive dependency that breaks other deps
     #       e.g `spacy`'s dependency `click` isn't installed
-    "typer!=0.26.0,!=0.26.1,!=0.26.2",
+    "typer!=0.26.*",
+    # transitive dependency that spacy needs but doesn't delcare...
```
jocelyneholdbrook: minor - type in declare. same below

jocelyneholdbrook (commented):
typo in declare otherwise looks good to me

jocelyneholdbrook (approved):

PR #519, opened 2026-05-29
test(medcat-trainer): Add simple example dataset for use for UAT testing

alhendrickson:
File will be used as part of UAT - this exact entry means that an annotation will be shown when using the medcat demo model in medcat-test-models.


mart-r (approved):
Fake it 'till you make it!

PR #522, opened 2026-06-01
Determine admin status based on Keycloak roles

jocelyneholdbrook:
- Implements role-based access control (RBAC) to identify if a user has admin privileges using Keycloak roles.
- Simplifies user role management within medcat-trainer.


alhendrickson (approved):
lgtm, approved!

For another day - you could support both? Instead of removing

Also alternatively could it be configurable? Might potentially make it a lot easier to deploy / explain. Not sure if apps like opensearch have hard coded roles but Id reckon they can be configured

I'd wanted to add pydantic-settings to trainer to replace the env vars parsing, but until then...

In the helm values I'd picture some yaml like this, then something to convert that to env vars.

```
auth:
   admin:
       roles:
        - medcattrainer_superuser
      groups:
        - medcattrainer-admins
```

PR #524, opened 2026-06-01
feat(medcat-trainer): trainer-ee-scaffolding

tomolopolis:
Scaffolding changes for the mct OSS project here according to the [plan](https://github.com/CogStack/cogstack-private/commit/194125cff938f7fcb4fe49be8c25cc03ef5e4fbd#diff-229356a32d6eb50d079751d5ca9ec7f9a09f4264c7a79448f7f38f0137dc9cf2) 

Plugin scaffolding changes for `TrainAnnotations` `ProjectAdmin` and security features in mct-ee

### DIFF
--- a/medcat-trainer/webapp/api/api/extensions.py
+++ b/medcat-trainer/webapp/api/api/extensions.py
@@ -1,0 +1,193 @@
+"""
+MedCAT Trainer extension API.
+
+Stable contract used by enterprise / third-party plugins. The OSS app itself
+emits these signals and consults the registries from a small number of
+explicit sites (see :mod:`api.signals`, :mod:`api.views`, :mod:`api.permissions`).
+
+Plugins are discovered via the ``mct.plugins`` Python entry-point group; each
+entry point points at a Django ``AppConfig`` subclass. Plugin URL configs are
+mounted at ``/api/ee/<app_label>/`` when the ``AppConfig`` sets
+``is_mct_plugin = True``. See :mod:`core.settings` and :mod:`core.urls`.
+
+The shape of this module is contract-tested by ``test_extensions.py`` and the
+``contract-tests`` CI job; changes that break the documented shape are
+breaking changes.
+"""
+from __future__ import annotations
+
+from collections.abc import Callable, Iterable
+from typing import Any, Optional
+
+from django.dispatch import Signal
+
+# ---------------------------------------------------------------------------
+# Signals
+# ---------------------------------------------------------------------------
+#
+# All signals MUST include at least the documented kwargs. Extra kwargs may
+# be added over time but documented kwargs are stable.
+
+#: Sent immediately before a document is submitted to training.
+#: kwargs: ``project`` (ProjectAnnotateEntities), ``document`` (Document),
+#:         ``user`` (User or None).
+pre_document_submit = Signal()
+
+#: Sent immediately after a document is submitted to training.
+#: kwargs: ``project``, ``document``, ``user``.
+post_document_submit = Signal()
+
+#: Sent after an :class:`AnnotatedEntity` row is created.
+#: kwargs: ``annotation`` (AnnotatedEntity), ``project``, ``document``,
+#:         ``user`` (or None).
+annotation_created = Signal()
+
+#: Sent after an :class:`AnnotatedEntity` row is updated.
+#: kwargs: ``annotation``, ``project``, ``document``, ``user``.
+annotation_updated = Signal()
+
+#: Sent after an :class:`AnnotatedEntity` row is deleted.
+#: kwargs: ``annotation`` (instance prior to delete), ``project``, ``document``.
+annotation_deleted = Signal()
+
+#: Sent after a :class:`ProjectGroup` row is created.
+#: kwargs: ``project_group``.
+project_group_created = Signal()
+
+#: Sent after a :class:`ProjectGroup` row is updated.
+#: kwargs: ``project_group``.
+project_group_updated = Signal()
+
+#: Sent after the OIDC user resolver returns a Django user.
+#: kwargs: ``user``, ``id_token`` (dict), ``created`` (bool).
+user_oidc_resolved = Signal()
+
+
+# ---------------------------------------------------------------------------
+# Permission hook registry
+# ---------------------------------------------------------------------------
+#
+# Permission hooks are grant-only: a hook returning ``True`` grants the
+# permission; ``None`` or ``False`` abstains and the OSS default decision is
+# used. Hooks MUST NOT be used to deny access that the OSS code would
+# otherwise grant.
+
+PermissionHook = Callable[..., Optional[bool]]
+_permission_hooks: dict[str, list[PermissionHook]] = {}
+
+
+def register_permission_hook(name: str, fn: PermissionHook) -> None:
+    """Register a grant-only permission hook for ``name``.
+
+    Hooks are called in registration order until one returns ``True``.
+    Any value other than ``True`` (including ``None`` and ``False``) abstains.
+    """
+    if not callable(fn):
+        raise TypeError("permission hook must be callable")
+    _permission_hooks.setdefault(name, []).append(fn)
+
+
+def get_permission_hooks(name: str) -> Iterable[PermissionHook]:
+    return tuple(_permission_hooks.get(name, ()))
+
+
+def clear_permission_hooks(name: Optional[str] = None) -> None:
+    """Clear hooks. Intended for tests."""
+    if name is None:
+        _permission_hooks.clear()
+    else:
+        _permission_hooks.pop(name, None)
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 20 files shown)


medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+# All signals MUST include at least the documented kwargs. Extra kwargs may
+# be added over time but documented kwargs are stable.
+
+#: Sent immediately before a document is submitted to training.
+#: kwargs: ``project`` (ProjectAnnotateEntities), ``document`` (Document),
+#:         ``user`` (User or None).
+pre_document_submit = Signal()
+
+#: Sent immediately after a document is submitted to training.
+#: kwargs: ``project``, ``document``, ``user``.
```
mart-r: I guess type information is consistent with the one above, but do we perhaps want to be explicit about it?

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+post_document_submit = Signal()
+
+#: Sent after an :class:`AnnotatedEntity` row is created.
+#: kwargs: ``annotation`` (AnnotatedEntity), ``project``, ``document``,
```
mart-r: Here as well, type information for `project` and `document` isn't included.

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+#:         ``user`` (or None).
+annotation_created = Signal()
+
+#: Sent after an :class:`AnnotatedEntity` row is updated.
+#: kwargs: ``annotation``, ``project``, ``document``, ``user``.
```
mart-r: Again, typing not included

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+annotation_updated = Signal()
+
+#: Sent after an :class:`AnnotatedEntity` row is deleted.
+#: kwargs: ``annotation`` (instance prior to delete), ``project``, ``document``.
```
mart-r: Typing not included

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+annotation_deleted = Signal()
+
+#: Sent after a :class:`ProjectGroup` row is created.
+#: kwargs: ``project_group``.
```
mart-r: Typing not included

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+project_group_created = Signal()
+
+#: Sent after a :class:`ProjectGroup` row is updated.
+#: kwargs: ``project_group``.
```
mart-r: Typing not included

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+project_group_updated = Signal()
+
+#: Sent after the OIDC user resolver returns a Django user.
+#: kwargs: ``user``, ``id_token`` (dict), ``created`` (bool).
```
mart-r: Typing not included (for first 2)

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+
+    ``item`` MUST contain ``id`` (str) and ``label`` (str). It SHOULD
+    contain ``route`` (str) or ``href`` (str). Additional keys pass through
+    verbatim to the frontend.
+    """
+    if not isinstance(item, dict):
+        raise TypeError("menu extension item must be a dict")
+    if "id" not in item or "label" not in item:
+        raise ValueError("menu extension item requires 'id' and 'label'")
+    _menu_extensions.append(dict(item))
```
mart-r: I suppose this is meant to pass a copy of the dict so changes to it somewhere else won't affect the registered extension. However, the shallow copy could still leave dict values (which could only be additional keys/values as per spec) vulnerable to outside changes.
Though probably not a massive issue, just something to keep in mind. If it's an issue `copy.deepcopy` could be used to avoid this.

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def get_menu_extensions() -> list[dict[str, Any]]:
+    return [dict(it) for it in _menu_extensions]
```
mart-r: Same note here regarding shallow copy

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+    """Register a frontend route descriptor exposed via ``GET /api/bootstrap/``.
+
+    ``route`` MUST contain ``path`` (str) and ``component`` (str — module
+    specifier or registered component name resolved by the frontend).
+    """
+    if not isinstance(route, dict):
+        raise TypeError("route must be a dict")
+    if "path" not in route or "component" not in route:
+        raise ValueError("route requires 'path' and 'component'")
+    _plugin_routes.append(dict(route))
```
mart-r: Same note here regarding shallow copy

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def get_routes() -> list[dict[str, Any]]:
+    return [dict(r) for r in _plugin_routes]
```
mart-r, line 152: Same note here regarding shallow copy

medcat-trainer/webapp/api/core/plugin_discovery.py
```diff
@@ -0,0 +1,75 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def discover_mct_plugin_app_configs() -> list[str]:
+    """Return dotted paths of :class:`~django.apps.AppConfig` classes to append to INSTALLED_APPS."""
+    try:
+        try:
+            eps = entry_points(group='mct.plugins')
+        except TypeError:
```
mart-r: We don't need this `except` clause. This would only be applicable for pre python 3.10:
https://docs.python.org/3/library/importlib.metadata.html#entry-points

3.9 (and below) are end of life, so I don't think we need to keep trying to support them.
Plus, we're running 3.12 for the image anyway.

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -0,0 +1,193 @@
...
+# ---------------------------------------------------------------------------
+# Permission hook registry
+# ---------------------------------------------------------------------------
+#
+# Permission hooks are grant-only: a hook returning ``True`` grants the
+# permission; ``None`` or ``False`` abstains and the OSS default decision is
+# used. Hooks MUST NOT be used to deny access that the OSS code would
+# otherwise grant.
+
+PermissionHook = Callable[..., Optional[bool]]
```
mart-r: Theser have an unknown argument signature. Is there a reason not to force args `User` and `ProjectAnnotateEntites`? That's what's being passed in `api.permissions.is_project_admin`, right? Or is that just because the method it's used in doesn't have typing? Perhaps at the very least define as 2 arguments as `[Any, Any]`?

mart-r (approved):
Overall looks good to me!

A few comments on type informations in extensions.

A question / suggestion regarding tyhe typing of permission hooks.

A note about shallow copies in extensions/plugins routes management.

A suggestion to simplify entry point discovery (no need to support python<3.9-style).

PS:
Didn't look at the frontend stuff really.

medcat-trainer/webapp/api/api/extensions.py
tomolopolis: good point - have typed and limited to the two args used

medcat-trainer/webapp/api/core/plugin_discovery.py
tomolopolis: done

medcat-trainer/webapp/api/api/extensions.py
tomolopolis: done

medcat-trainer/webapp/api/api/extensions.py
tomolopolis: done

medcat-trainer/webapp/api/api/permissions.py
```diff
@@ -14,7 +14,7 @@ def has_permission(self, request, view):
         return request.method in permissions.SAFE_METHODS
 
 
-def is_project_admin(user, project):
+def is_project_admin(user: [User], project: [ProjectAnnotateEntities]):
```
mart-r, line 17: I don't think this is correct type annotation?

tomolopolis: ah - yep - thanks

PR #525, opened 2026-06-01
build(medcat-scripts): Fix issue from mistyping of pandas index.

mart-r:
This PR is fixing an issue that's surfaced by a  update (3.0.3.260530) where the  is interpreted as  rather than  by . For  the  method would expect an argument, but for  it does not. We know this works in practice since there are multiple tests that call this method explicitly.

For reference, the  issue surfaced in workflow:
```
dquote bquote> evaluate_mct_export/mct_analysis.py:230: error: Missing positional argument "level" in call to "droplevel" of "Index"  [call-arg]
```


tomolopolis (approved):
lgtm

PR #526, opened 2026-06-03
paper(medcat): MedCAT 2 paper related scripts and documentation

mart-r:


alhendrickson (commented):
Minor one from me, could you make a README.md in the `medcat-v2/paper` folder? Super high level one that basically says `run pip install somewhere, then get the required data following each data folder, then ./run_all_at_once.sh` or whatever.

Just something that explains /paper folder from a dev perspective with this repo open, I get the paper itself will explain a lot.

tomolopolis:
looks good for paper result reproduction. 
Can you confirm:
- none of this is included in the install of the lib? 
- should any of these script runs be run in a GHA build?

mart-r:
> * none of this is included in the install of the lib?

Goot point! As of now, it would have. But I've added a `MANIFEST.in` to avoid that.

> * should any of these script runs be run in a GHA build?

That isn't really possible. Not only is there no availability of models that this would work (full models, embedding linker model) for or the data (MIMIC raw data, linking datasets) that this uses, even if it did, the entire thing might take longer than the 6h allocated to job runners.

tomolopolis (approved):
yeah agreed - worth adding in the README what version of the lib that this was last tested with. 
lgtm otherwise

PR #527, opened 2026-06-04
bug(medcat): CU-869djf7qd Fix trainer detected name preprocessing

mart-r:
This PR fixes the preprocessing of names during supervised training.

The issue was that for names which had tokens that would normally need to be skipped (e.g new lines) would not be skipped during the `prepare_name` call. That's because these skips are annotated in the tagger.
E.g (in an example from Adam) the annotation for:
```
electrolytes
were monitored
```
would previously be processed into:
```
electrolytes~\n~were~monitored
```

This PR uses the `Pipeline.tokenizer_with_tag` property instead of just the `.tokenizer`. The former is a tokenizer that also runs the tagger component.

With this change, the same name is correctly processed to:
```
electrolytes~were~monitored
```

There are also a few tests to this effect.

### DIFF
--- a/medcat-v2/medcat/trainer.py
+++ b/medcat-v2/medcat/trainer.py
@@ -408,6 +408,19 @@ class Trainer:
                     docs, current_document, train_from_false_positives,
                     devalue_others)
 
+    def _get_processed_name(self, raw_name: str) -> str:
+        pn_dict = prepare_name(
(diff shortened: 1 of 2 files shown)


adam-sutton-1992 (approved):
Yeah nice, lucky to find that issue.

PR #530, opened 2026-06-10
chore(medcat-trainer): Harden MCT plugin scaffold with trust docs and URL/signal guards.

tomolopolis:
Document the in-process plugin trust model, validate bootstrap menu/route URLs against dangerous schemes, isolate plugin signal failures from core flows, and add secure API authoring guidance.

### DIFF
--- a/medcat-trainer/webapp/api/api/extensions.py
+++ b/medcat-trainer/webapp/api/api/extensions.py
@@ -13,14 +13,36 @@ mounted at ``/api/ee/<app_label>/`` when the ``AppConfig`` sets
 The shape of this module is contract-tested by ``test_extensions.py`` and the
 ``contract-tests`` CI job; changes that break the documented shape are
 breaking changes.
+
+Security / trust model
+----------------------
+An installed ``mct.plugins`` package runs as **first-class Django application
+code**: it executes at import time and in ``AppConfig.ready()`` with full access
+to the database, filesystem, environment, and request/session data. There is no
+sandbox. Treat plugins like kernel modules — only install packages you trust and
+have vetted. See ``docs/plugins.md`` for the full trust model and secure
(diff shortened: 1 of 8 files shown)


alhendrickson (approved):
Looks good to me

medcat-trainer/webapp/api/api/extensions.py
```diff
@@ -67,6 +89,74 @@
...
+                "MCT plugin signal receiver %r raised %s: %s",
+                getattr(receiver, "__qualname__", receiver),
+                type(response).__name__,
+                response,
+                exc_info=response,
+            )
+
+
+# ---------------------------------------------------------------------------
+# URL hardening for bootstrap-exposed entries
```
alhendrickson: Only comment is I'd probably remove any mention of "security" or "hardening" throughout, and really just talk about it being a safety net for writing plugins. You've already got docs that say it isn't security, so would keep that message consistent 

Would do a pass through of the documentation/code to reword like that really. 

For example this code here doesnt give any security or hardening at all for malicious plugins, (and I get that you acknowledge that elsewhere ). I'd just write a plugin that edits this extensions.py file and restart trainer.

alhendrickson: For the example

Given this pydoc:
```
The registry helpers below apply light input hardening (URL-scheme validation on
menu/route entries) and signal emission is isolated via :func:`dispatch` so that
a buggy or hostile plugin receiver cannot break core request flows. These are
defence-in-depth measures, **not** a security boundary against malicious code
running in-process.
```

I'd probably remove the "hardening", "defence-in-depth"  and "hostile plugin receiver" text here. Keep it consistent that it isn't defence at all. Function is called "validate", which feels like the right word to use throughout

PR #532, opened 2026-06-10
fix(medcat-trainer): Fix Concepts tab for CDBs with no heirarchy

alhendrickson:
### Existing Behavior
When used with a CDB that isn't a full SNOMED one, the concepts tab currently fails silently for the user. The network call gives a 400 "Requested MedCAT CDB model does not include parent2child metadata to explore a concept hierarchy"

image

This change fixes it by supporting a flat tree under a dummy root node. It just returns the first 500 concepts, a limit hardcoded for now

### New Behavior
image

### DIFF
--- a/medcat-trainer/webapp/api/api/views.py
+++ b/medcat-trainer/webapp/api/api/views.py
@@ -1064,19 +1064,28 @@ def view_metrics(request, report_id):
         return Response(200)
 
 
+FLAT_CDB_ROOT_CUI = 'Root'
+FLAT_CDB_CHILDREN_LIMIT = 500
+
+
 @api_view(http_method_names=['GET'])
 def cdb_cui_children(request, cdb_id):
     parent_cui = request.GET.get('parent_cui')
     cdb = get_cached_cdb(cdb_id, CDB_MAP)
 
-    # root SNOMED CT code: 138875005
-    # root UMLS code: CUI:
-
-    if cdb.addl_info.get('pt2ch') is None:
-        return HttpResponseBadRequest('Requested MedCAT CDB model does not include parent2child metadata to'
-                                      ' explore a concept hierarchy')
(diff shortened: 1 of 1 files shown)


medcat-trainer/webapp/api/api/views.py
```diff
@@ -1064,19 +1064,28 @@ def view_metrics(request, report_id):
...
-    if cdb.addl_info.get('pt2ch') is None:
-        return HttpResponseBadRequest('Requested MedCAT CDB model does not include parent2child metadata to'
-                                      ' explore a concept hierarchy')
+    if not cdb.addl_info.get('pt2ch'):
+        if parent_cui is None:
+            return Response({'results': [
+                {'cui': FLAT_CDB_ROOT_CUI, 'pretty_name': 'All concepts'},
+            ]})
+        cuis = sorted(cdb.cui2info)[:FLAT_CDB_CHILDREN_LIMIT]
+        return Response({'results': [
```
tomolopolis, line 1082: Can we 'warn' that the model does not allow ontology navigation due to pt2ch map metadata missing?  and / or this has been limited to CHILDREN_LIMIT in the resp? 

The whole idea of the concepts tab was an easy way to build out a filter for a given project. Which this still allows over just throwing an error so does make sense.

tomolopolis (approved):
lgtm - thanks for fixing the obv issues with this screen. a 'warning' / notice or even disabling this tab gracefully for those cdb's with no pt2ch map likely makes sense

medcat-trainer/webapp/api/api/views.py
alhendrickson: Would agree some UI side warning would be really useful. In general as well we could do with more toasts/errors etc shown around 

Made a ticket on https://app.clickup.com/t/20516280/869dnb54u

PR #534, opened 2026-06-11
test(medcat-trainer): Fix the Concept view test broken in PR 532

alhendrickson:
Fixes the test broken by this one https://github.com/CogStack/cogstack-nlp/pull/532

Excitingly this was overlooked due to the timing of the branches, my branch didn't actually have this test to break... should all be great in future though now we have a lot more tests and coverage


tomolopolis (approved):
lgtm

PR #535, opened 2026-06-11
docs(medcat-service): Fix link formatting and add v2 model pack section

alhendrickson:
Fixed the link formatting for the example-medcat-v1-model-pack and added a new section for the example-medcat-v2-model-pack.


tomolopolis (approved):
lgtm

PR #537, opened 2026-06-11
Fix/medcat trainer/error handling fetch entities

tomolopolis:
Fixes an error where documents / entities were not appearing despite having a valid auth token. A `fetchData` silent failure was masking a session error that was fixed with a session refresh.


alhendrickson (approved):
lgtm - I was actually getting some 401s today so this looks well timed

PR #539, opened 2026-06-12
feat(medcat-trainer): scaffold for sourcing models via a  medcattery instance

tomolopolis:
small change to allow plugins to hook into model import signal


alhendrickson (approved):
lgtm

Future wildcard would be to use medcat-den for it potentially. Personal idea for future would be if we can make it download the model on demand instead of requiring it to be setup on disk, this would let it scale horizontally (so much more other stuff required to do that as well though)

tomolopolis:
indeed - `medcat-den` will be the way forward, but using the `medcattery` backend. So should be relatively straightforward on the `mct-ee` side of things.

PR #541, opened 2026-06-15
chore(medcat): Change log WARN to DEBUG for tokenizer internals path

alhendrickson:
Change this warning to debug. It being a warning seems to be misleading.

Triggers a lot on any test models, including the ones in the test models folder.

 Eg can see this warning in `medcat-v2-tutorials/notebooks/introductory/basic/4._Evaluating_performance_on_dataset.ipynb`

```
Loading a model with incorrectly saved tokenizer internals path. Was saved as 'models/base_model/en_core_web_md' whereas should have had just 'en_core_web_md'. This is an automated fix - no further action is needed
```


tomolopolis (approved):
lgtm

PR #543, opened 2026-06-15
feat(medcat-trainer):  Set token expiry for OIDC sessions

jocelyneholdbrook:
## Summary
This Pull Request implements a token refreshing mechanism for OIDC (OpenID Connect) sessions in the MedCATTrainer client and introduces corresponding unit tests to validate the feature.
- Added _refresh_oidc_token and ensure_token_fresh methods for handling OIDC token refreshes and ensuring token validity.
- Updated existing authentication code to integrate with the new refreshing mechanism, including storing token expiry.
- Introduced unit tests for verifying different scenarios of the OIDC token refresh logic (test_ensure_token_fresh suite).
- Adjusted relevant test setups for proper mocking and validation of token refresh behavior.
This change improves authentication session management and adds robustness to OIDC workflows.

### DIFF
--- a/medcat-trainer/client/mctclient.py
+++ b/medcat-trainer/client/mctclient.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from datetime import datetime
 import json
 import os
+import time
 from abc import ABC
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -287,6 +288,9 @@ class MedCATTrainerSession:
         self.username = username or os.getenv("MCTRAINER_USERNAME")
         self.password = password or os.getenv("MCTRAINER_PASSWORD")
         self.server = server or 'http://localhost:8001'
+        self._keycloak_settings: Optional[KeycloakSettings] = None
+        # positive infinity so that token refresh is never triggered for non-OIDC sessions
+        self._token_expiry: float = float('inf')
 
         env_use_oidc = os.getenv("MCTRAINER_USE_OIDC", "")
         env_use_oidc_truthy = env_use_oidc.strip() == "1"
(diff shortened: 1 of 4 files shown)


medcat-trainer/client/mctclient.py
```diff
@@ -314,6 +317,18 @@ def __init__(self, server=None, username=None, password=None, keycloak_settings=
         else:
             raise MCTUtilsException(f"Failed to login to MedCATtrainer instance running at: {self.server}")
 
+    def _refresh_oidc_token(self) -> None:
+        """Fetch a new OIDC access token and update the Authorization header."""
+        token = get_keycloak_access_token(self._keycloak_settings)
+        self.headers = {"Authorization": f"Bearer {token}"}
+        # Refresh 60s before the typical 5-minute Keycloak access token lifetime (i.e. 4 minutes from now)
+        self._token_expiry = time.monotonic() + 240
```
alhendrickson: Minor - would probably let this be an env var, can default to the 4 mins still

medcat-trainer/client/mctclient.py
```diff
@@ -314,6 +317,18 @@ def __init__(self, server=None, username=None, password=None, keycloak_settings=
...
+
+    def ensure_token_fresh(self) -> None:
```
alhendrickson, line 328: Does this need to be wired in somewhere? Wondering how this is triggered

alhendrickson (approved):

medcat-trainer/client/mctclient.py
jocelyneholdbrook: it can be called by whoever uses this client. The idea is that you check whether the current OIDC token is about to expire, and if so, fetches a new one from Keycloak before the next API call goes out. I will be using it in medcattery's medcat-trainer.py. But in an ideal scenario it should be checked in every medcat-trainer HTTP request but I wanted to avoid making a load of changes at this stage.

medcat-trainer/client/mctclient.py
jocelyneholdbrook: done

PR #545, opened 2026-06-15
ci: set the version number before building the package in medcat-trainer_ci

jocelyneholdbrook:
### Summary

- Ensure version number is set before building the package to avoid upload issues with duplicate version numbers.
- Prevents from silently failing to upload a package with the same version number each time


alhendrickson (approved):
lgtm

PR #550, opened 2026-06-16
feat(transformer_ner): Trainable BERT NER plugin, along with rawstring tokenizer that splits on all spaces.

adam-sutton-1992:
Hihi,

This is WIP so we'll top it off with a TODO here for now:

- Unit Testing for transformer_NER - DONE!
- Unit Testing for rawstring_tokenizer - DONE!
- Testing all variants of components that are viable (aka all except RawStringTokenizer + ContextBasedLinking) -  DONE!
- Parameter testing  - DONE!
- README for transformer_ner  - DONE!
- README for rawstring_tokenizer - DONE!
- Whatever mypy and linting fails I missed :D -  DONE!

____________________________________________________________________________________________________

This is the trainable MLM transformer model attempting to do NER. It is a binary BIOES NER model where each prediction is either (Beginning-Ent, Inside-Ent, Outside-Ent, End-Ent, Single-Ent). This is a bit of an advancement compared to BIO models, where E signals the end of a multi token label, and S signals a stand alone token label. We try to prioritise B and E tokens here for performance (i.e. ensure we get well formed predictions). We also have a CRF head after the MLM model to try to encourge more well formed label predictions (i.e. only I and E after B

`transformer_ner.py` is the main logic for the plugin, while `transformer_ner_model` is the logic for the model (such as initialisation, loading, and the forward step). 

Rawstring_tokenizer is a tokenizer where all tokens are based on whitespace splits i.e. new lines, tabs, and spaces. It still can't perfectly obtain all entities (sub word entities). But is an improvement for where some entities don't have spacy representation. This mainly improves performances in pipelines where it's using transformer_ner, and embedding_linker.

There are also additional changes to the embedding_linker. I understand these should probably be seperate, however that's slipped through the cracks. Apologies. The changes are mainly more functionality and configurability:

1. Multiple entities per span
2. Short and long similarity thresholds, along with top_k entities being passed
3. Appending the pre_inference link candidates to those from vocab based methods (which are both a part of transfomer_ner and vocab_ner).
4. Additional documentation in the code for all of this step by step.

Performance wise you can expect performances of trained models with _reasonable_ configs to look like this (based on training / testing of Distemist & Snomed Entity Linking Benchmark):

1. Spacy Tokenizer + Vocab based NER + Context Based Linker: Recall 0.7
2. Spacy Tokenizer + Vocab based NER + Embedding Linking: Recall 0.75
3. Spacy Tokenizer + Transformer based NER + Context Based Linking: Recall 0.73
4. Rawstring Tokenizer + Transformer based NER + Embedding Linking: Recall 0.85

There a few additional pieces with these metrics. The embedding linker is highly configurable so you can go from a Recall of 0.84-ish and a Precision of 0.4, to 0.9 recall and 0.05 precision. These metrics I have here are essentially on configurations I think make "sense". One such measure of that metric is "if the recall goes up, and precision remains the same improves I'd consider that a solid improvement".  I have documented these changes in performance within the config, so hopefully people can make informed decisions.


adam-sutton-1992:
Hi, just a brief mention on performances. I'll list some permutations of components, tokenisers and just list what I can get. These aren't exhaustive and the performance can always be improved with better configs. 

EDIT:

For posterity testing performances was done with the RawString Tokenizer - however the regex tokenizer performs better and does the same task. So assume these performances can be _very slightly_ better.

- Baseline (with training): 0.68 Recall
- Baseline (without training): 0.34 Recall

- Rawstring Tokenizer, <Any NER>, Context Based Linker (CBL): Not possible!
- Rawstring Tokenizer, TRF NER, Trained Embedding Linker (TEL): 0.85 Recall 
- Rawstring Tokenizer, TRF NER, Static Embedding Linker (SEL): 0.54 Recall

- SpacyTokenizer, TRF NER, SEL: 0.48 Recall
- SpacyTokenizer, TRF NER, TEL: 0.7 Recall
- SpacyTokenizer, TRF NER, CBL: 0.69 Recall
- SpacyTokenizer, Vocab NER, SEL: 0.35 Recall
- SpacyTokenizer, Vocab NER, TEL: 0.75 Recall

mart-r:
> Hi, just a brief mention on performances. I'll list some permutations of components, tokenisers and just list what I can get. These aren't exhaustive and the performance can always be improved with better configs.
> 
> * Baseline (with training): 0.68 Recall
> * Baseline (without training): 0.34 Recall
> * Rawstring Tokenizer, , Context Based Linker (CBL): Not possible!
> * Rawstring Tokenizer, TRF NER, Trained Embedding Linker (TEL): 0.85 Recall
> * Rawstring Tokenizer, TRF NER, Static Embedding Linker (SEL): 0.54 Recall
> * SpacyTokenizer, TRF NER, SEL: 0.48 Recall
> * SpacyTokenizer, TRF NER, TEL: 0.7 Recall
> * SpacyTokenizer, TRF NER, CBL: 0.69 Recall
> * SpacyTokenizer, Vocab NER, SEL: 0.35 Recall
> * SpacyTokenizer, Vocab NER, TEL: 0.75 Recall

As much as I agere that recall is often more important than precision, on it's own I don't see it as particularly valuable. I could just mark all spans for all concepts and get a recall of 1 at the expense of zero precision.
So maybe add some idea of precision as well?
Also, how long do these take to run? Is there any throughput changes?
And what dataset are you running this on?

PS:
Will look at the PR itself later on, just saw this as a separate thread on Gmail...

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
```diff
@@ -0,0 +1,253 @@
...
+               base_char_index + match.start(), 
+               base_char_index + match.end())
+
+class Token:
+    def __init__(self, 
+                 text: str, 
+                 index: int, 
+                 char_index: int, 
+                 end_char_index: int) -> None:
+        # --- BaseToken fields ---
```
mart-r: This is fine, but to avoid all the boilerplate, I would just use the `.text` (and so on) fields rather than the property stuff below.

It would still satisfy the protocol. The only difference is that the user doesn't know that these are writable, and thus type checkers would complain if/when try try to write into these.

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
```diff
@@ -0,0 +1,253 @@
...
+    def norm(self) -> str: return self._norm
+    @norm.setter
+    def norm(self, val: str) -> None: self._norm = val
+    
+class Entity:
+    _addon_extension_paths: set[str] = set()
+
+    def __init__(self, text: str, start_index: int, end_index: int,
+                 start_char: int, end_char: int, label: str = "") -> None:
+        # --- BaseEntity fields ---
```
mart-r: Same here - I would juse use fields instead of properties.

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
```diff
@@ -0,0 +1,253 @@
...
+import unicodedata
+import re
+
+
+# keep both hyphens and slashes within words
+_WORD_RE = re.compile(r"[^\W_]+(?:[^\W_]+)*", re.UNICODE)
+# _WORD_RE = re.compile(r"[^\W_]+(?:[-/][^\W_]+)*", re.UNICODE)
+
+
+def _iter_word_spans(
```
mart-r: This is effectively where the tokenization itself happens :) 

Perhaps move this (along with the regex above) to its own module to separate it? i.e `base.py`

medcat-plugins/rawstring-tokenizer/README.md
```diff
@@ -0,0 +1,80 @@
...
+    make_archive=False,
+)
+print("Saved model pack to:", f"{TARGET_FOLDER.rstrip('/')}/{TARGET_PACK_NAME}")
+```
+
+## How It Works
+
+### Component Registration
+
+Register the tokenizer by name before trying to add the tokenizer to the pipeline. If loading a model with a rawstring tokenizer register it beforehand.
```
mart-r: I'd normally expect the extension to take care of that.
See:
https://github.com/CogStack/cogstack-nlp/blob/main/medcat-plugins/embedding-linker/src/medcat_embedding_linker/__init__.py
and 
https://github.com/CogStack/cogstack-nlp/blob/main/medcat-plugins/embedding-linker/src/medcat_embedding_linker/registration.py
for example

EDIT:
I see you've already done that for transformers NER as well

medcat-plugins/rawstring-tokenizer/README.md
```diff
@@ -0,0 +1,80 @@
+# MedCAT Embedding Linker
```
mart-r: How does it differ from the regex tokenizer?
Just a different regex?
What's the benefit?
Why do we need this separately from the built in regex tokenizer?

Also, would be great to add a workflow that runs some typing/linting and tests on this stuff.

medcat-plugins/rawstring-tokenizer/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Component Registration
+
+Register the tokenizer by name before trying to add the tokenizer to the pipeline. If loading a model with a rawstring tokenizer register it beforehand.
+
+### Embedding Generation
+
+## Limitations
+
+- Can NOT be used with the default `context_based_linker` as, that uses spacy tokens and spacy embeddings for linking. Which are not used with this tokenizer.
```
mart-r: I'm not quite sure I understand why it can't be used with the context based linker.
There is nothing in there that is coupled to the spacy tokenizer.

medcat-plugins/transformer-ner/README.md
```diff
@@ -0,0 +1,100 @@
+# MedCAT Embedding Linker
```
mart-r: This could also use workflows that run linting/typing/tests.

mart-r (changes requested):
Massive PR!
This should have been 3 different PRs:
1. Embedding linker changes
2. Rawstring tokenizer
3. Transformers based NER

I really couldn't go through everything in enough detail at this point.
But I've left a few comments.

I think the main questions I had were the following:
1. What's rawstring tokenizer and how does it differ from the built in regex tokenizer? Why do we need a separate one?
2. Why can't the rawstring tokenizer be used with the regular / context based linker?

And then there's the fact that I think we need to have workflows for the new things as well.

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
adam-sutton-1992: I can do that - but I think they're somewhat required?

I have to manually shift token indexes back when using `entity_from_tokens`, or `entity_from_tokens_within_doc`.

medcat-plugins/rawstring-tokenizer/README.md
adam-sutton-1992: yeah sorry it is done. Probably a hold over from being in the other components documentation. I'll remove them.

medcat-plugins/rawstring-tokenizer/README.md
adam-sutton-1992: It can! My bad. Fundamentally though that context was embedding from w2v embeddings each time.

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
mart-r: If you need it somewhere, go for it.

But my point was that for mose of these you're just returning the `_`-prefixed value. So it's easier (i.e less code) to not separate the attribtue and the property.

I.e instead of 
```python
# in __init__
        self._text = text
# later down
    @property
    def text(self) -> str: return self._text
```
you can just do
```python
# in __init__
        self.text = text
```
and not have to deal with the property.

medcat-plugins/rawstring-tokenizer/README.md
adam-sutton-1992: Hi - yeah the Rawstring Tokenizer is redundant because the regex does the job perfectly. With a slight performance increase. At least this didn't hurt development time. I'll take out the RawString stuff. Thank you.

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
adam-sutton-1992: I was aiming for consistent design here for all attributes / properties.

If you prefer it to be as you suggest here I will do so.

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
adam-sutton-1992: As Above!

medcat-plugins/rawstring-tokenizer/src/medcat_rawstring_tokenizer/tokens.py
mart-r: I'm fine either way - I just personally prefer the version with less code to maintain.

medcat-plugins/transformer-ner/README.md
mart-r: This says embedding linker, but it's transformers NER!

medcat-plugins/embedding-linker/src/medcat_embedding_linker/embedding_linker.py
```diff
@@ -569,28 +736,24 @@ def _pre_inference(
         le: list[MutableEntity] = []
         to_infer: list[MutableEntity] = []
         for entity in all_ents:
+            # if no candidates just skip it
+            if not entity.link_candidates:
+                continue
+            # TODO: Check if this is right now with multiple entites being possible
```
mart-r: Did you end up checking? Fine if out of scope.

medcat-plugins/transformer-ner/src/medcat_transformer_ner/transformer_ner.py
```diff
@@ -0,0 +1,580 @@
...
+    def train(self, cui: str,
+            entity: MutableEntity,
+            doc: MutableDocument,
+            negative: bool = False,
+            names: Union[list[str], dict] = []) -> None:
+        """Train the NER component on a given document. This is used in the
+        supervised training loop of the MedCAT trainer.
+        """
+        # if this is the last entity, we'll train
+        # kind of a hacky work around, but it's minimal impact on the CAT trainer
```
mart-r, line 238: I don't quite like the way this works.
The method _could_ be called for a subset of the entities in text. But you're always using all of them.

I would prefer is this cached the entities sent to it.

PS:
That does still fall over if the last entity is skipped for whatever reason. So maybe keep track of the doc and do old doc stuff if different doc is sent?

medcat-plugins/transformer-ner/tests/resources/mct2_model_pack.zip
mart-r, line 1: Is this the same model that's in `medcat-test-models/`? Maybe you can just refer to that? Though this would - of course - mean that the tests wouldn't be portable (i.e for source distributions) but I don't think it's a massive issue?

medcat-plugins/transformer-ner/pyproject.toml
```diff
@@ -0,0 +1,124 @@
...
+requires-python = ">=3.10"
+
+license = {text = "Apache-2.0"}
+
+keywords = ["ML", "NLP", "NER+L"]
+
+authors = [
+  {name = "A. Sutton"},
+  {name = "T. Searle"},
+  {name = "M. Ratas"},
```
mart-r, line 19: I'm fine not being listed - I didn't write any of this :) 
But up to you

medcat-plugins/transformer-ner/pyproject.toml
```diff
@@ -0,0 +1,124 @@
...
+maintainers = [
+  {name = "CogStack", email = "contact@cogstack.org" }
+]
+
+classifiers = [
+  # How mature is this project? Common values are
+  #   3 - Alpha
+  #   4 - Beta
+  #   5 - Production/Stable
+  "Development Status :: 3 - Alpha",
```
mart-r, line 34: We just need to make sure to remember to go non-alpha if/when we're ready. I'm pretty sure PyPI picks up this classifier and displays it.

mart-r (changes requested):
A few comments and a docs change.

Overall looks good to me. Really close!

medcat-plugins/transformer-ner/README.md
adam-sutton-1992: Done!

adam-sutton-1992: God I swear I fixed this like 5 times.

medcat-plugins/embedding-linker/src/medcat_embedding_linker/embedding_linker.py
adam-sutton-1992: Yeah no both todos in the linker have been done/confirmed - just not removed.

medcat-plugins/transformer-ner/src/medcat_transformer_ner/transformer_ner.py
adam-sutton-1992: We had this problem with the linker - the workaround is there was to call the `_train_on_batch` method manually - for samples that wasn't a full linker batch size.

But in normal training for the embedding linker we also have the same process:

```
if (
            len(self.training_batch) >= self.cnf_l.training_batch_size 
            or entity is doc.linked_ents[-1]
        ):
```

Realistically I think that the larger "better" change would be to have training done on a per document basis in each component that trains. Of course this impacts the core library.

I can't see in the current training setup where you wouldn't train on any entity in a provided document, and we don't expose a way to do so.

mart-r: From the context of what this is doing, I think the current method signature makes a lot of sense. All it is, is a user saying "this entity within this context - that's the ground truth". Anything else that's annotated within the document shouldn't generally be relevant. The only reason that it's relevant for you is that otherwise it becomes very slow due to running in series.

In essence, we _are_ training on each individual entity rather than a collection of entities in a document. I.e this isn't designed to be able to learn which entities appear together in a document - that's a downstream task.

You're right that it's not currently used in isolation anywhere. But it's still well within the API that someone _could_ do that.

I think you're right that it's extremely unlikely. But I'd still like _some_ protection against this since it's not something that's explicitly encouraged / blocked / disallowed. So perhaps just keep track of the latest document? I.e default to `None` and reset to `None` after training the doc; set it to the document if/when in this method, but if it's already set to a different document, raise an exception since you don't know what to do with it.

medcat-plugins/transformer-ner/tests/resources/mct2_model_pack.zip
adam-sutton-1992: This is consistent with embedding linker.

I can change both if you prefer :D

medcat-plugins/transformer-ner/pyproject.toml
adam-sutton-1992: Still p much an author for review / input

medcat-plugins/transformer-ner/pyproject.toml
adam-sutton-1992: Is there a time you'd swap between alpha -> beta? Maybe when submitting to places?

medcat-plugins/transformer-ner/tests/resources/mct2_model_pack.zip
mart-r: Yeah, but that's from before I moved them central :) 
So yeah, if it works for you, change both.

medcat-plugins/transformer-ner/pyproject.toml
mart-r: At some point you'd probably want to move to production/stable is my point :)

medcat-plugins/transformer-ner/src/medcat_transformer_ner/transformer_ner.py
adam-sutton-1992: > You're right that it's not currently used in isolation anywhere. But it's still well within the API that someone could do that.

I'd think that's a long logical leap to protect from.

Train is called and exposed, subsequent calls are "protected" meaning they shouldn't _really_ be called from outside the library. If you're defensively programming for situations where users do that then I'd think it'd be fair (and better) to say "use it properly or fork". 

> In essence, we are training on each individual entity rather than a collection of entities in a document. I.e this isn't designed to be able to learn which entities appear together in a document - that's a downstream task.

That's what all models are doing, no? No models from my knowledge are using the co-occurrence of entities as part of it's training / prediction. From an engineering perspective it just makes more sense for almost all models to train on all entities at once. If this was a context free embedding / LSTM solution it would still be doing the same. If it was a generative model it would be doing the same (with much more serious computational consequences if multiple calls were inefficiently made to a train method). Passing the same document multiple times to a component multiple times sounds somewhat redundant from a software perspective (although of course it has no real impact on performance). 

I'm happy to keep a track of the latest document - and then encourage users to manually call a train method (consistent with the embedding linker, besides that having potentially multiple documents entities). I personally think it's "more" hacky than was is there currently.

adam-sutton-1992: After speaking online, we've agreed to keep it as is.

adam-sutton-1992:
Okay - think I've done all the changes as requested. Apologies for the massive PR.

mart-r (approved):
Looks good to me

PR #551, opened 2026-06-17
ci(medcat-service): Change GHA to build and test the image before push

alhendrickson:
I think this fixes another PR is why I'm doing it now

Change is to follow this pattern, make a "test" image that is local only, test it, then push it to dockerhub afterwards

https://docs.docker.com/build/ci/github-actions/test-before-push/

This pattern already used in other builds


jocelyneholdbrook (approved):
Lgtm

PR #542, opened 2026-06-15
fix(medcat): Relcat addon fix deserialise_from to match metacat behavior

alhendrickson:
## Error

When I save a model pack in medcat trainer, that has a relcat addon, then it errors the page and returns 500

The error looks like the RelCAT addon isn't as safe as the Metacat addon.

This change updates the relcat deserialise_from method to match the metacat one, by adding defaults and some safety into it. It should look like `MetaCATAddon`

### Trainer

```


2026-06-15 16:01:27.406 | [medcattrainer] INFO 2026-06-15 15:01:27,406 cat.py l:784:Unziping the model pack and loading models.
2026-06-15 16:01:31.444 | [medcattrainer] [15/Jun/2026 15:01:31] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:01:39.503 | [medcattrainer] [15/Jun/2026 15:01:39] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:01:45.148 | [medcattrainer] INFO 2026-06-15 15:01:45,147 serialisers.py l:314:Determined serialised of type dill off disk
2026-06-15 16:01:47.901 | [medcattrainer] [15/Jun/2026 15:01:47] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:01:55.852 | [medcattrainer] [15/Jun/2026 15:01:55] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:02:03.842 | [medcattrainer] [15/Jun/2026 15:02:03] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:02:11.930 | [medcattrainer] [15/Jun/2026 15:02:11] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:02:24.381 | [medcattrainer] [15/Jun/2026 15:02:24] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:02:28.315 | [medcattrainer] [15/Jun/2026 15:02:28] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
2026-06-15 16:02:36.492 | [medcattrainer] [15/Jun/2026 15:02:36] "GET /api/prep-docs-bg-tasks/ HTTP/1.1" 200 36
[...]

PR #552, opened 2026-06-17
chore(medcat-trainer): Update medcat version to 2.8.3

alhendrickson:
This is to fix the bug uploading relcat models

https://github.com/CogStack/cogstack-nlp/pull/542


dcstang (approved):
lgtm

PR #553, opened 2026-06-18
hotfix(medcat): Disable config train at load time

mart-r:
This PR sets `config.components.linking.train` to `False` at load time.

Some models may have been saved with this set to `True`. And as of v2.8 (specifically #414) this can cause issues since that method of running training is no longer used.

There's also an added test that makes sure the loaded model has this config option disabled. (NOTE: the model that's in the repo has this problem so the test previously set this to `False` itself, but I've now removed this from the test class setup).

I also checked that this test fails in the main branch (i.e without the change from the first commit).

PS:
Just to be clear, this has to do with the config option that was previously used for self-supervised training. The process was to set `*.train = True` and run inference and then the vocab based linker would perform training within the inference step. The expected new approach is more explicit and does not use this config option. **So all models can still be trained** - both unsupervised and supervised.

### DIFF
--- a/medcat-v2/medcat/cat.py
+++ b/medcat-v2/medcat/cat.py
@@ -125,9 +125,9 @@ class CAT(AbstractSerialisable):
         """
         # pass
         if self.config.components.linking.train:
-            logger.info("Training was enabled during %s. "
-                        "It was automatically disabled.",
-                        stage)
+            logger.debug("Training was enabled during %s. "
+                         "It was automatically disabled.",
+                         stage)
             self.config.components.linking.train = False
 
     @overload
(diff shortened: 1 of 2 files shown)


tomolopolis (approved):
lgtm

medcat-v2/medcat/cat.py
```diff
@@ -125,8 +125,9 @@ def _ensure_not_training(self) -> None:
         """
         # pass
         if self.config.components.linking.train:
-            logger.warning("Training was enabled during inference. "
-                           "It was automatically disabled.")
+            logger.info("Training was enabled during %s. "
+                        "It was automatically disabled.",
```
alhendrickson: Minor - I'd probably move this to debug? 

Like right now using some of the distributed models I think this will be displayed each time and generally confuse the logs a bit. Had complaints about some similar log lines that always triggered.

mart-r: Agreed - especially now that this is triggered during load rather than during inference AND that it doesn't have much of an actual effect.

I.e before #414 this could be a genuine issue if not dealt with during inference. But now that this isn't actually used and is only a legacy config option, we really don't need to treat it as important to the user.

PR #554, opened 2026-06-18
feature(medcat): Allow filtering addons upon model load

mart-r:
This is in response to #542 .

The target for this method (`CAT.load_addons`) is trainer and it's (currently) only loading MetaCAT addons.

So I've amended the method such that it is able to only load specfic addons if/when required.

PS:
There's a side effect of this where the addon type is now a class level attribute rather than an instance level one. But effectively that should be the case anyway.

NOTE:
The trainer changes themselves can't be included before the core lib changes are merged in and published because trainer runs against the latest (pinned) releases and not the current state of the core library.

### DIFF
--- a/medcat-v2/medcat/components/addons/addons.py
+++ b/medcat-v2/medcat/components/addons/addons.py
@@ -13,11 +13,9 @@ class AddonComponent(BaseComponent, Protocol):
     """Base/abstract addon component class."""
     NAME_PREFIX: str = "addon_"
     NAME_SPLITTER: str = "."
+    # NOTE: need to implement
+    addon_type: str
     config: ComponentConfig
-
-    @property
-    def addon_type(self) -> str:
-        pass
 
     def is_core(self) -> bool:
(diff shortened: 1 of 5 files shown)


medcat-v2/medcat/components/addons/addons.py
```diff
@@ -13,12 +13,10 @@ class AddonComponent(BaseComponent, Protocol):
     """Base/abstract addon component class."""
     NAME_PREFIX: str = "addon_"
     NAME_SPLITTER: str = "."
+    # NOTE: need to implement
```
alhendrickson: Could you add more to this comment in the code? I read this like it's a todo, but I dont think it is?

alhendrickson (approved):

medcat-v2/medcat/components/addons/addons.py
mart-r: Yep, agreed

PR #555, opened 2026-06-22
test(medcat-trainer): Add model inference test with config-train set to true

mart-r:
This PR adds the test that _should_ be fixed in core lib PR #553 .
However, this is still likely to fail right now since there has been no release of medcat with the change included.


alhendrickson (approved):
Epic, this is like the real test of the app. Thanks

mart-r:
> Epic, this is like the real test of the app. Thanks

Real test, for realzies.

PR #558, opened 2026-06-23
fix(medcat-trainer): Fix issue with meta-annotation tasks disappearing

mart-r:
Full details available here:
https://discourse.cogstack.org/t/manual-meta-task-selection-overwritten-when-using-model-pack-without-metacat-models/403

But the TLDR is that in 2.6.1 the `instanceof` fix in #529 (the `signals.py` change, specifically - the branch within it would previously never be reached) made this always trigger the use of tasks from the meta cats in the model pack and never default to the manual ones set for the project.

First commit adds a test, next one adds the relevant fix.


alhendrickson (approved):
lgtm with the test

As said maybe we can revisit the signals in general

PR #562, opened 2026-06-23
chore(medcat-trainer): Update medcat version to 2.8.6 in pyproject.toml

alhendrickson:


mart-r (approved):
If/when the workflows are happy, who am I to complain?

PR #565, opened 2026-06-24
fix(medcat-trainer): Dont use (most) addons in the pipeline for document processing

mart-r:
Refer to #564 
This just adds / fixes a few minor things.

The real fix needs to happen on the core lib side at a later date.

PS:
Allowed `meta_cat` addons at load time since they're actually used / expected in some bits.


tomolopolis (approved):
lgtm

PR #566, opened 2026-06-24
feat(medcat): Allow loading with fewer addons

mart-r:
This PR allows loading a subet of addons at model load time.

This should be beneficial in (at least) 2 situations (both in terms of load time and RAM usage):
1. You've got a model with addons you do not wish to use
2. You've got a model with addons you cannot use (i.e you don't have the external code required)

PS:
In medcat terms, `addons` are components that run in addition to the core pipe (NER, linking) and `extensions` are external projects that may provide core components and addons alike (as long as they register them with the core lib).

### DIFF
--- a/medcat-v2/tests/test_cat.py
+++ b/medcat-v2/tests/test_cat.py
@@ -558,6 +558,8 @@ class CatWithMetaCATSaveLoadTests(CatWithMetaCATTests):
     def test_filter_meta_cat_loads_meta_cat(self):
         loaded = cat.CAT.load_model_pack(self.mpp, keep_addons_of_types=[MetaCATAddon])
         self.assert_has_one_meta_cat(list(enumerate(loaded.pipe.iter_addons())))
+        mc = next(loaded.pipe.iter_addons())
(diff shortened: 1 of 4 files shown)


medcat-v2/tests/test_cat.py
```diff
@@ -545,11 +545,20 @@ def test_can_load_meta_cat(self):
...
         self.assertEqual(len(addons), 1)
         _, addon = addons[0]
         self.assertIsInstance(addon, MetaCATAddon)
 
+    def test_filter_non_existant_loads_no_addons(self):
+        from medcat.components.addons.relation_extraction.rel_cat import RelCATAddon
+        loaded = cat.CAT.load_model_pack(self.mpp, keep_addons_of_types=[RelCATAddon])
+        self.assertFalse(list(enumerate(loaded.pipe.iter_addons())))
+
+    def test_filter_meta_cat_loads_meta_cat(self):
```
tomolopolis, line 558: worth adding an assert to test the correct metacat by name has been preserved?

tomolopolis (approved):
lgtm

PR #567, opened 2026-06-24
feat(medcat-trainer): Avoid loading meta cat upon model pack upload

mart-r:
This PR allows model packs with MetaCATs (and other addons) to be uploaded without the need to load all these addons.

The reason they were previously loaded (though notably addons other than MetaCAT weren't used) was to get some information for meta tasks.

In an ideal world, we'd have all the sufficient information saved in the model card. But alas, we don't.

So what this PR does is load the model pack's config (only) and gets the relevant information from there.

This should save a considerable amount of time (and memory) since loading all these addons (in an ephemral way, no less) can consume a lot of resources and loading just the (admittedly dill-serialised) json should be a lot leaner.

**NB! Limitations:**
This is currently hard-coded to `dill` serialisations of the model pack. While this is currently the only serialisation we provide, this does create tight coupling. So we may need to review this in the future.


medcat-trainer/webapp/api/api/tests/test_with_dummy_model.py
tomolopolis, line 1: generally dislike the term 'fake' model. either use `dummy` or `example`? fake personally sounds like its not a test or perhaps not a model at all, when in fact it is both a real model and even a real model on disk by the looks of it.

medcat-trainer/webapp/api/api/tests/test_with_fake_model.py
```diff
@@ -108,3 +114,57 @@ def test_can_use_model_for_inference(self):
...
+        # A real user — the view reads request.user
+        self.user = create_user(username="testuser", password="password", is_staff=True)
+
+    # NOTE: mocking dispatch just so it won't complain about missing sender
+    @patch('api.model_import.dispatch')
+    def test_model_import_does_not_load_addons(self, mock_dispatch):
+        with patch.object(MetaCATAddon, "__init__") as mock_init:
+            import_model_pack(
+                MODEL_PATH,
+                name='fake-test-model',
```
tomolopolis: just test-model is fine

alhendrickson (approved):
lgtm, should really speed it up

tomolopolis (approved):
small comments

tomolopolis:
Lgtm

PR #572, opened 2026-06-30
feat(medcat-trainer): Speed up project upload

mart-r:
This PR attempts to speed up project upload.

I was having trouble uploading the entity linking challenge project (i.e conversion from the raw linking challenge) to the app.cogstack.org trainer. It seemed to just crash it.
However, when I tried on the demo trainer (sites.er.kcl.ac.uk), it worked just fine.

So I thought perhaps the app.cogstack.org is timing out.

Looking at the timings, it looked like it was taking around 5m30s to upload the dataset to the the trainer instance I was running locally.
I got the time down to around half.
More specifically, it took 328.675s before, and with the changes it took 150.610s.

Here's the changes:
- Avoid resaving the ProjectAnnotateEntities twice
  - The signal caused every save to happen twice for the entire DB entry
- Avoid some file system IO by caching CUI filter file
  - This was also loaded off disk at every save time
- Use model update instead of save for modifcation date of project annotate entities
  - The idea being that instead of updating all fields, only the relevant field gets updated

EDIT:
The next step would be to run this as a background task. But then some of the UI would need to also reflect that. So didn't get a chance to do that yet.

### DIFF
--- a/medcat-trainer/webapp/api/api/models.py
+++ b/medcat-trainer/webapp/api/api/models.py
@@ -354,7 +354,7 @@ class Relation(models.Model):
         return str(self.label)
 
 
-def update_project_last_modified(project: ProjectAnnotateEntities, last_modified):
+def update_project_last_modified(project: 'ProjectAnnotateEntities', last_modified):
     # so that the local project last_modified is updated
     project.last_modified = last_modified
     # so that the project last_modified is updated in the database
--- a/medcat-trainer/webapp/api/api/models.py
+++ b/medcat-trainer/webapp/api/api/models.py
@@ -354,6 +354,14 @@ class Relation(models.Model):
         return str(self.label)
 
 
+def update_project_last_modified(project: ProjectAnnotateEntities, last_modified):
+    # so that the local project last_modified is updated
+    project.last_modified = last_modified
+    # so that the project last_modified is updated in the database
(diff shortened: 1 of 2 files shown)


alhendrickson (approved):
Lgtm, approved

Though high level if this is literally saving some text into a db it should be ms not minutes, though maybe something can get redesigned somewhere at a later date. One difference between the app.cogstack.org one is that it's running postgres (potentially on a different VM) vs local using sqlite, so if there's a lot of db calls it might be a real issue there.

mart-r:
> Lgtm, approved
> 
> Though high level if this is literally saving some text into a db it should be ms not minutes, though maybe something can get redesigned somewhere at a later date. One difference between the app.cogstack.org one is that it's running postgres (potentially on a different VM) vs local using sqlite, so if there's a lot of db calls it might be a real issue there.

Yeah, totall agreed. My expectation was that It'd be easy to get this down to a few seconds. But turned out it wasn't.

I'm pretty certain the underlying issue is the constant IO. I.e within this process, each entity is saved separately. And upon saving each entity, the project data (last modified) is also changed. So there's a lot of IO when doing sqlite. I did try to wrap it in atomic to try and force it to write once, but that turned out to be slower for some reason.

In any case, potentially some work for another day. But I can leave the latest timings log in here for future reference:
https://gist.github.com/mart-r/b6498f301bf6fd186b41b983abfe204e

PR #579, opened 2026-07-15
fix(medcat-trainer): Secret_key env var is now used instead of defaulting

alhendrickson:
## Behavior before

If I set env to PROD, and provide a SECRET_KEY env var, it was never used, django still loaded the hardcoded one

## Behavior after
If I provide a SECRET_KEY env var, it is now used (no matter the environment)


mart-r (approved):
Good fix!

Workflow should be fine after #580 is in.

PR #581, opened 2026-07-16
fix(medcat-trainer): Avoid creating duplicate MetaCAT models upon model upload

mart-r:
When uploading a model with meta cats that have the same name as ones that already exist on trianer, the current implementation tries to re-register the meta cat with the same name. But that fails due to the duplicate name.

This PR addresses that by changing the name when it already exists and saves the meta-cats with a unique name in such cases.

The idea is to avoid trying to save the duplicate while allowing for similarly named (potentially) different meta cats.

### DIFF
--- a/medcat-trainer/webapp/api/api/models.py
+++ b/medcat-trainer/webapp/api/api/models.py
@@ -122,8 +122,11 @@ class ModelPack(models.Model):
             for meta_cat_dir, meta_cat_cnf in meta_cat_addons:
                 meta_model_name = f'{meta_cat_cnf.general.category_name} - {meta_cat_cnf.model.model_name}'
                 if MetaCATModel.objects.filter(name=meta_model_name).exists():
-                    logger.info("Meta CAT Model '%s' already exists, avoiding duplicate registration", meta_model_name)
-                    continue
+                    prev_meta_can_name = meta_model_name
+                    meta_model_name = f'{meta_model_name} - from {self.id}'
+                    logger.info(
+                        "Meta CAT Model '%s' already exists, avoiding duplicate registration, renaming to %s",
+                        prev_meta_can_name, meta_model_name)
                 mc_model = MetaCATModel()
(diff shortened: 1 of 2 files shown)


medcat-trainer/webapp/api/api/models.py
```diff
@@ -120,9 +120,16 @@ def save(self, *args, skip_load=False, **kwargs):
             meta_cat_addons = load_meta_cat_info_from_model_folder(unpacked_model_pack_path)
             metaCATmodels = []
             for meta_cat_dir, meta_cat_cnf in meta_cat_addons:
+                meta_model_name = f'{meta_cat_cnf.general.category_name} - {meta_cat_cnf.model.model_name}'
+                if MetaCATModel.objects.filter(name=meta_model_name).exists():
```
alhendrickson: As a thought - could you just always set it to `f'{meta_model_name} - from {self.id}'`, and not bother with the exists check? Would just be faster/simpler.

```
mc_model.name = f'{meta_cat_cnf.general.category_name} - {meta_cat_cnf.model.model_name} - {self.id}'
```

I'd assume this doesnt break anything preexisting as its already saved but would check. 

Alternatively, is there any reason to have the unique index on name at all? I'm thinking if there is a reason (eg we lookup by name somewhere) then this might break here. I reckon there's always an ID field generated and used for the joins, so potentially we can just remove the unique index? I do like how few changes there are here so approved if it's all good.

alhendrickson (approved):
Approved, with loose thoughts before

medcat-trainer/webapp/api/api/models.py
mart-r: I did think of just always doing the id-based name. And - as far as I can tell - this name isn't recreated anywhere from scratch. So I _think_ this should be fine indeed.

Plus, if we always use the ID then we wouldn't need to do another DB query to check for existance.

alhendrickson (approved):

PR #582, opened 2026-07-16
feat(medcat): Add component contracting

mart-r:
This PR adds component contracting.

So far, what each component needs and provides has only been implicitly defined in the existing component behaviour. But it has not really been defined anywhere.

This PR adds such capabilities to the library. More specifically, it adds per core component type contracting. This contract isn't checked at run time, but the PR also adds simple helper methods that make verifying contracts extremely easy.

For instance, the simplest way to check that all the components (for now, only NER and linker) would be:
```python
from unittest import TestCase
from medcat.components.contracting_testing import assert_component_contracts
# implement create_model_with_my_component
class MyComponentTest(TestCase):
    def test_my_model_contract(self):
        # create or load a model with your custom component(s)
        # NOTE: This would (generally) need to be able to NER / link 1 entity in the example text
        #       The test time models in medcat would be sufficient
        cat = create_model_with_my_component()
        assert_component_contracts(cat)
```

What the contracting tests check:
- Whether all the expected must-provide fields are written (per document or token)
- Whether all the expected needs fields are read (per document or token)
- Whether all the output entities have the required fields (e.g `detected_name` for NER and `cui` for linker)

Notably, the utility method also makes a few assumption. For instance, it assumes that for the NER step `token.to_skip` is called for each token (which it normally is) and `ner_ents` are written once. And for the linker, it assumes that `ner_ents` are read once and `linked_ents` written once. Plus, it assumes that the output of either has at least 1 entity (otherwise collection contracting cannot be checked).

**Open question**
Should we allow enabling this for "the first document" in production with a (defaults-to-false) config option?


alhendrickson:
Could you add this into the documentation too? As you've basically already written it in the PR description which explains it well

mart-r:
> Could you add this into the documentation too? As you've basically already written it in the PR description which explains it well

I tried to figure out where to put this and went on the docs page:
https://docs.cogstack.org/projects/nlp/en/latest/
It's a mess between actual v2 things and legacy v1 things. For instance, the quick start example code block straight up won't work on v2.
I'll address that in a separate PR.

For this one, I'll just add some bits to the README.

medcat-v2/README.md
```diff
@@ -1,3 +1,218 @@
 # MedCAT v2
 
-MedCAT v2 is now simply in at (../medcat)[./medcat].
\ No newline at end of file
+<<<<<<< HEAD
```
tomolopolis: remove merge indicator

medcat-v2/README.md
```diff
@@ -1,3 +1,218 @@
...
+  volume=117,
+  pages="102083",
+  month=jul,
+  year=2021,
+  issn="0933-3657",
+  doi="10.1016/j.artmed.2021.102083"
+}
+</details>
+```
+=======
```
tomolopolis: and here

medcat/medcat/pipeline/pipeline.py
```diff
@@ -341,8 +341,33 @@ def get_doc(self, text: str) -> MutableDocument:
         Returns:
             MutableDocument: The resulting document.
         """
+        return self.pipe_until(text, None)
+
+    def pipe_until(
```
tomolopolis, line 346: this is nice little feature - can you add to docs / guide update when updating with comp contracting side of things?

tomolopolis (approved):
lgtm

medcat-v2/README.md
mart-r: Most of this file should be removed, really - it's in `medcat/docs/index.md` now. So did that.

medcat/medcat/pipeline/pipeline.py
mart-r: Done now

PR #584, opened 2026-07-20
docs(medcat): Fix docs

mart-r:
Currently the docs:
https://docs.cogstack.org/projects/nlp/en/latest/
Are quite different to the README:
https://github.com/CogStack/cogstack-nlp/tree/main/medcat-v2

And in fact, historically, these have been two (almost) identical files we track.
But right now, the `index.md` was well behind the actual state and the `README.md` in this project. In fact, some of it was plain wrong.

This PR tries to use one markdown file for both of them.

PS:
Also added reference to PyPI version in README (already existed in the docs).

PPS:
Also added back v1 news in a separate markdown file.


alhendrickson:
Hey I get the plan, but I think I'd suggest the alternative:

I'd probably prefer to keep both files, but instead strip like 95% of the README.md file out and just say "see docs [here]" or something.

For the reason, I think we will need the index.md to become mkdocs specific. For example we should (later) fill out the docs to be more user friendly, so I'd see the need for an index.md that at the end says "next click here for architecture [relative path]", which wouldnt work in a README.md file at all. Basically I think we will need to include stuff that only makes sense when it's hosted on a site, like diagrams, links and formatting etc, we can use the features that are available.

Similarly I'd imagine some stuff that makes sense in the README.md that doesnt make sense for being on a website, like local dev steps that use local files etc.

mart-r:
> Hey I get the plan, but I think I'd suggest the alternative:
> 
> I'd probably prefer to keep both files, but instead strip like 95% of the README.md file out and just say "see docs [here]" or something.
> 
> For the reason, I think we will need the index.md to become mkdocs specific. For example we should (later) fill out the docs to be more user friendly, so I'd see the need for an index.md that at the end says "next click here for architecture [relative path]", which wouldnt work in a README.md file at all. Basically I think we will need to include stuff that only makes sense when it's hosted on a site, like diagrams, links and formatting etc, we can use the features that are available.
> 
> Similarly I'd imagine some stuff that makes sense in the README.md that doesnt make sense for being on a website, like local dev steps that use local files etc.

The current implementation does keep the `index.md` (turned out I couldn't link to an out-of-scope README). It just pulls the contents of the README and displays them there.

I think if/when we discover that we want to only keep a part of the README in the docs, we can split the relevant parts out. I.e either push out things that don't belong in the docs into other markdown files (and refer to them in README), or keep the docs-related info in a separate file and link to it in the README (and in `index.md`).

My main concern here is keeping the same information in multiple places. The current state of `index.md` clearly shows that only one of them will be updated (for the most part, at least).

alhendrickson (approved):
Meant to hit approve the other day, apologies

I'm all good for this to go in as obviously much better than it used to be. Would want to revisit later on though!

PR #589, opened 2026-07-24
build(cogstack-es): Pin ruff to earlier versions to avoid workflow failures

mart-r:
Turns out ruff `v0.16.x` introduced new checks that I do not wish to tackle at this point.

E.g:
- I001 [*] Import block is un-sorted or un-formatted
- UP035 [*] Import from `collections.abc` instead: `Iterable`, `Sequence`
- UP035 `typing.Type` is deprecated, use `type` instead
- UP007 [*] Use `X | Y` for type annotations
- UP045 [*] Use `X | None` for type annotations

### DIFF
--- a/cogstack-es/pyproject.toml
+++ b/cogstack-es/pyproject.toml
@@ -34,7 +34,9 @@ dev = [
     "pandas-stubs",
     "types-tqdm",
     "pytest",
-    "ruff",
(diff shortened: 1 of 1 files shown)


tomolopolis (approved):
lgtm - fair

PR #591, opened 2026-07-26
fix(medcat): Fix config defaulting to train=True.

mart-r:
This default would normally be fixed at model load time or at inference time (see calls to ). However, when building a new model (e.g with a mostly default config), the value would remain . And if there is no save/load or inference before training, this would fail to train because of it.

PS: To reiterate, this would not have affected any users down the line since there's fixes at model load time. This would have only affected new model creation (with default/unchanged config).


tomolopolis (commented):
lgtm

PR #595, opened 2026-07-28
feat(medcat-service): Demo app supports custom footers for display

alhendrickson:
## Current behavior
- We start up the medcat-service app, open the demo UI and it is hardcoded to show either the footer for medcat demo, or anoncat demo.
- These are the right thing to show for the hosted demo apps we have, but are misleading for any other deployment.
- For example, for some hospital running this service in production for anoncat, if they open the UI they always see this string "Please DO NOT test with any real sensitive PHI data." which isn't true for them.

## New Feature
- You can now optionally pass in a path to a markdown file, and it will show that instead
- A sensible default has been left in place that just links to this repo

## Next step in other repository
- Update the hosted demo apps with the previous footers, so they show the right thing again in that context


mart-r (approved):
Couldn't the demo-time disclaimer still be included? But just default to the new / default one?

That way, all that is needed for the demo is a reference to this.

But if it's easier to have that closer to where the demo is deployed, then I'm fine with that as well.

Overall, looks good! And will surely avoid some confusion in the future.

alhendrickson:
> Couldn't the demo-time disclaimer still be included? But just default to the new / default one?
> 
> That way, all that is needed for the demo is a reference to this.
> 
> But if it's easier to have that closer to where the demo is deployed, then I'm fine with that as well.

For sure, hopefully it will be easier to do from the deployment side, essentially will be able to just put the desired contents in the config there.

We could look to add it back here, though I'm thinking that the original files were actually super specific to our hosted demo deployment instances, so it doesnt really make sense to have in the public docker images in a way. Definitely the Deid one at least which was like "these are the concepts in this specific model" doesn't make sense as it runs any model.

PR #598, opened 2026-08-12
feat(medcat): Allow bulk superivsed training

mart-r:
This PR allows bulk supervised training.

The idea is that instead of the previous supervised training method:
```
def train(cui: str, entity: MutableEntity, doc: MutableDocument, negative: bool, names: dict) -> None
```
we now use a new one:
```
def train_supervised_batch(self, examples: list[TrainingExample]) -> None
```
This refers to the method on the component wishing to be trained in a supervised manner.

## The changes
There's a fair few changes that had to be made to make this all work

### New protocol for supervised training

The new protocol is the `BatchTrainableComponent`.
This only defines the new method.
Any future implementation is expected to use this going forward.

The method uses `TrainingExample` objects.
This is a dataclass that keeps track of information relevant to each example.
This includes stuff like the CUI, the entity, the document, whether it's a negative example, how many epochs to do.
But on top of that, it also keeps track of things (empty for now, but these were also empty before) for name description, status, and ontology (which can be used when adding new names).

### Supporting both the old and new method

Because this would otherwise be a breaking change, I wanted to support both the old and the new method.
This means that components that follow the old protocol should still work.
The main workhorse for this is the `_LegacyBatchAdapter` in `type_utils`.
This adapts the old style to the new one by calling the old method in sequence.

### Avoiding changing names for supervised training

There's a part of the vector context model that allows the (artificial) changing of the name of the concept being trained.
At some chance (as per config), the name that actually appears in the text can be replaced with a synonym.
This was meant to include less-common synonyms in self-supervised training.
However, for supervised training, this does not make a lot of sense.
As such, this was disabled.

I would say that this was somewhat of a bug before.
I don't think it should have ever affected supervised training.

### Fixed some annotation/entity mismatch issue

Since the PR #374 (in v2.7.0) there has been an issue with annotation / antity mismatch when doing supervised training.
The culprit was the preprocessing of each annotation that meant that some (ones that weren't able to be mapped to an entity - i.e sub-token annotations) were dropped.
And when iterating over the entities and annotations later (in a zip), after the dropped annotation there will have been mismatches there.

This has now been fixed.

### Some changes to string ordering

Due to the random nature of string hashing, getting ordered output from sets (e.g when taking a random sample for name changes) is unpredictable even if the underlying random number generator is seeded.
This can be mitigated by setting the `PYTHONHASHSEED` environmental variable.
But I didn't think the current situation was desirable - if someone seeds the RNG, they expect to get the same output run to run.

So to mitigate this, in some places (only in the name-change part I believe), string sets are sorted before picking a random item from them.

### Some extra debug output

While trying to figure out the differences, I added some extra debug output.
Some of this can be removed if we decide to.
But it helped me identify the various issues I faced along the way.

## The testing I did to make sure this still works as close to the previous version as possible:

In order to verify that the updated training method did (roughly speaking) the same thing, I ran a number of tests locally.
I ran the tests for two full models (the 2023 model and the 2025 model).
I ran it over a small number of documents from the entity linking challenge.
I ran it over a number of different configurations:
- With `devalue_others=True/False`
  -  This was affected by the changes so was keen to do the test with and without
- With `train_from_false_positives=True/False`
  - Same as above
- With full models and traininig-cleared models

During the tests I seeded the `PYTHONHASHSEED` to avoid string-hashing issues.

As the output I used:
- Basic info (`CDB.get_basic_info`)
- Per CUI count train (`CDB.get_cui2count_train`)
- Per name count train (`CDB.get_name2count_train`)
- Per CUI context vectors

What I observed was that in all cases the basic information (i.e the first 4 in the above list) were identical.
While the context vectors did change in some scenarios (more on the reasons below), the differences were not huge.

### Small tutorials change

Added a part for these protocols to the tutorial on creating your own components here:
https://github.com/CogStack/cogstack-nlp/blob/feat/medcat/allow-bulk-superivsed-training/medcat-v2-tutorials/notebooks/advanced/2._Create_and_use_component.ipynb

## The overall caveats:

There were some differences between the two models were observed in the cases where `devalue_others=True` was set.
The reason for these differences is boils down to the order of operations.
Previously, the devaluation of others (i.e training on negative examples of other concepts that share the same name to push these further from the current context) was done on a per annotation basis directly after a specific annotation was trained.
However, with the bulk training setup, they are done in bulk after the entire batch has been trained.
This can cause issues due to randomness being involved (e.g for negative sampling), but also due to simply training the same concept / name in a different order (i.e if it had a positive annotation and was considered a negative example due to `devalue_others` in another annotation in the same document).

Overall I am relatively happy with the small caveat.
But it is still important to note.

### DIFF
--- a/medcat-v2/medcat/components/linking/vector_context_model.py
+++ b/medcat-v2/medcat/components/linking/vector_context_model.py
@@ -54,6 +54,7 @@ class ContextModel(AbstractSerialisable):
         self.name_separator = name_separator
         self._disamb_preprocessors = (  # copy if default/empty
             disamb_preprocessors or disamb_preprocessors.copy())
+        self._in_supervised_training = False
 
     def get_context_tokens(self, entity: MutableEntity, doc: MutableDocument,
                            size: int,
@@ -107,6 +108,8 @@ class ContextModel(AbstractSerialisable):
                 yield vec * self.weighted_average_function(step)
 
     def _should_change_name(self, cui: str) -> bool:
+        if self._in_supervised_training:
+            return False
         target = self.config.random_replacement_unsupervised
         if random.random() <= target:
             return False
@@ -340,8 +343,12 @@ class ContextModel(AbstractSerialisable):
             logger.warning("The provided entity for cui <%s> was empty, "
                            "nothing to train", cui)
             return
-        vectors = self.get_context_vectors(
-            entity, doc, per_doc_valid_token_cache, cui=cui)
+        self._in_supervised_training = True
+        try:
+            vectors = self.get_context_vectors(
+                entity, doc, per_doc_valid_token_cache, cui=cui)
+        finally:
+            self._in_supervised_training = False
         cui_info = self.cui2info[cui]
         lr = get_lr_linking(self.config, cui_info['count_train'])
         if not cui_info['context_vectors']:
--- a/medcat-v2/medcat/components/linking/vector_context_model.py
+++ b/medcat-v2/medcat/components/linking/vector_context_model.py
@@ -96,8 +96,8 @@ class ContextModel(AbstractSerialisable):
         return tokens_left, tokens_center, tokens_right
 
     def _tokens2vecs(self, tokens: Sequence[Union[MutableToken, str]],
-                    step_start: int = 0
-                    ) -> Iterable[np.ndarray]:
+                     step_start: int = 0
+                     ) -> Iterable[np.ndarray]:
         for step, tkn in enumerate(tokens, start=step_start):
             lower = tkn.lower() if isinstance(tkn, str) else tkn.base.lower
             if lower not in self.vocab:
@@ -116,7 +116,10 @@ class ContextModel(AbstractSerialisable):
                                   tokens_center: list[MutableToken]
                                   ) -> Iterable[np.ndarray]:
         if cui is not None and self._should_change_name(cui):
-            new_name: str = random.choice(list(self.cui2info[cui]['names']))
+            new_name: str = random.choice(
+                sorted(list(self.cui2info[cui]['names'])))
+            logger.debug("CUI %s changes name from '%s' to '%s'",
+                         cui, tokens_center[0].base.text, new_name)
             new_tokens_center = new_name.split(self.name_separator)
             return self._tokens2vecs(new_tokens_center)
         else:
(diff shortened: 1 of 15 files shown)


tomolopolis:
Are you updating the docs / guides of how supervised training should now be done?

tomolopolis:
code changes look fine, but the change needs docs / tutorial changes

mart-r:
> Are you updating the docs / guides of how supervised training should now be done?

Shouldn't actually affect end users at all. The underlying API (through the `trainer` object) remains the same.

But I do think I need to change some of the tutorials alongside this to make them work.

mart-r:
> > Are you updating the docs / guides of how supervised training should now be done?
> 
> Shouldn't actually affect end users at all. The underlying API (through the `trainer` object) remains the same.
> 
> But I do think I need to change some of the tutorials alongside this to make them work.

There wasn't really a place where anything needed to be changed as far as I could tell.

But I added a few more parts to the tutorial that deals with creation of custom components that now also go through the protocols needed for training.
The changed notebook is available here:
https://github.com/CogStack/cogstack-nlp/blob/feat/medcat/allow-bulk-superivsed-training/medcat-v2-tutorials/notebooks/advanced/2._Create_and_use_component.ipynb

medcat-v2/medcat/components/linking/vector_context_model.py
```diff
@@ -96,8 +97,8 @@ def get_context_tokens(self, entity: MutableEntity, doc: MutableDocument,
         return tokens_left, tokens_center, tokens_right
 
     def _tokens2vecs(self, tokens: Sequence[Union[MutableToken, str]],
-                    step_start: int = 0
-                    ) -> Iterable[np.ndarray]:
+                     step_start: int = 0
```
adam-sutton-1992, line 100: This is just spacing isn't it? I can't see a difference otherwise.

adam-sutton-1992 (approved):
LGTM.

While I think this is a "good" change on the law of averages...

I think that maintaining backward compatibility might be a bit of a hassle in the future. It might be worth hurting current development now than pushing it down the road. But that's your choice. Just a minor opinion.

adam-sutton-1992:
One final comment sorry - could this change affect how the plug-in components are called?

medcat-v2/medcat/components/linking/vector_context_model.py
mart-r: Yeah, just didn't like uneven look of the indentation and it was close to things I was fixing anyway :)

mart-r:
> One final comment sorry - could this change affect how the plug-in components are called?

Not sure what you mean...

adam-sutton-1992:
I mean does it require development to handle the entire batch, rather than one sample (label) at a time?

mart-r:
> I mean does it require development to handle the entire batch, rather than one sample (label) at a time?

No. That's the entire idea of doing the backwards compatibility. This still works for the old as well as the new API.

PR #601, opened 2026-08-14
feat(medcat-tutorials): Add custom addon tutorial

mart-r:
This PR adds a totorial regarding custom addons.

This should make it easier for someone to create one in the way that they were intended to be used.

The notebook can be seen here:
https://github.com/CogStack/cogstack-nlp/blob/feat/medcat-tutorials/add-custom-addon-tutorial/medcat-v2-tutorials/notebooks/advanced/3._Create_and_use_a_custom_addon.ipynb


tomolopolis:
This still leaves a little too much to the imagination? 
How about a v small LLM prompt to return a binary Yes / No this term is negated?

mart-r:
> This still leaves a little too much to the imagination? How about a v small LLM prompt to return a binary Yes / No this term is negated?

I think you're missing the point here.

This is a general custom addon tutorial rather than a specific LLM use case one (which is what we discussed yesterday - this PR was from Friday).

tomolopolis (approved):
lgtm

PR #602, opened 2026-08-14
feat(medcat): Avoid misconfigured components

mart-r:
This PR makes sure that the pipe does not tolerate components that return `None`.

This should help people who implement their own components better understand why their component might be silently  breaking the entire pipe.
Up until now, if the last component in the pipe returned `None`, this would propagate as the return value from `CAT.__call__` or result in empty dict output from `CAT.get_entities()`.

This can easily happen if/when creating a new component and forgetting the `return` statement in the `__call__` method.

NOTE:
Previously, if a component that was NOT the last component in the pipe, it would likely still break the pipe, but that would be due to the following component expecting a `MutableDocument` rather than `None`. So even in these cases the exception is now much clearer on the issue.


tomolopolis (approved):
lgtm

PR #606, opened 2026-08-18
feat(medcat): Get Stats: harder, better, (probably not) faster, stronger

adam-sutton-1992:
Hihi,

A new world for get_stats.

Adds three new character metrics:

- Character IoU (which is a pretty standard definition of a metric). 
- Golden Character IoU. Which is I believe an interesting metric for us. Essentially the character IoU of all CUIs that have a label. Something of a relation to the recall - I'd say.
- Cohen's Kappa. Used often for measuring inter-annotator agreement. However can be used as a metric of classificiation.

Adds two new "modes":

- "Perfect" linking. Effectively a measure of the performance of the NER step. This doesn't use a perfect linker. It just hacks all labels and predictions to have the same fake cui "__NER__".
- "Perfect" ner. A measure on the linking step.  This uses a the NER component that cheats, and thus is tasked with linking only.
- Obviously the full pipeline is also in there.

Minor changes:

- Returns a StatsCalculator object that can be explored for per cui and overall metrics, along with raw stats.
- Also counts the occasions where no tokens are found for a label span. This results in a false negative, as the model cannot possibly predict that entity. But it also counts the number of these, a decent metric for the tokenizer.
- I've also tested and adapted KFold so the metrics are identical to what they were previously. I've made minimal structural changes during testing and they all passed. 
- I hard coded testing for the Stats, and hand calculated what metrics came out should.

I'm a bit unhappy with the naming of the pydantic structure of "RawStats", "ProjectStats", "ModeStats"... It kind of makes sense but is a bit sloppy when reusing it.


mart-r:
NOTE:
Docs build failures are because of my work on #607 - in order to build there, I need to change the reference to `.readthedocs.yaml` in settings on RTD side. And this bricks everything other than my PR. I wanted to only change it temporarily (for now) but because there were issues with the docs build I kept it on the PR-specific path for a while. Which is why it failed here.
I should be able to change it back and rebuild the docs with the correct (for this PR) path.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+from pydantic import BaseModel, Field
+
+class RawStats(BaseModel):
+    """Raw accumulated state for a single evaluation mode."""
+
+    tp: int = 0
+    fp: int = 0
+    fn: int = 0
+    no_tokens: int = 0
+
```
mart-r, line 37: These could use some doc strings or something. Right now it's not clear what they are for the casual user.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+    char_cohen_k_n: float = 0.0
+
+
+class Metrics(BaseModel):
+    """Calculated metrics for a single evaluation mode."""
+
+    overall: OverallMetrics
+    per_cui: dict[str, CUIMetrics] = Field(default_factory=dict)
+
+class ModeStats(BaseModel):
```
mart-r, line 139: I think this could use a method that produces the same output as the previous iteration, i.e (to borrow from the other bits of code that already do this):
```python
def to_legacy(self) -> tuple:
    per_cui = (
                full_stats.metrics.per_cui
                if full_stats.metrics is not None
                else {}
            )
    return (
                self.stats.cui_fp,
                self.stats.cui_fn,
                self.stats.cui_tp,
                {cui: metrics.precision for cui, metrics in per_cui.items()},
                {cui: metrics.recall for cui, metrics in per_cui.items()},
                {cui: metrics.f1 for cui, metrics in per_cui.items()},
                self.stats.cui_gold_counts,
                self.stats.examples,
            )
```

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+class ProjectStats(BaseModel):
+    """Accumulated state and calculated metrics for one project 
+    or all projects."""
+    full_pipeline: ModeStats = Field(
+        default_factory=ModeStats
+    )
+    ner: ModeStats | None = None
+    linking: ModeStats | None = None
+
+    _MODE_FIELDS = {
```
mart-r: This feels like an `Enum`?
Right now it's just some magic strings hidden somewhere.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+    """Accumulated state and calculated metrics for all projects."""
+    all_projects: ProjectStats = Field(
+        default_factory=ProjectStats
+    )
+    projects: dict[int, ProjectStats] = Field(
+        default_factory=dict
+    )
+    
+    
+    def get_projects(self, project_index: int = -1) -> list[ProjectStats]:
```
mart-r: This feels a little weird. -1 usually refers to the last element. But here it's "all" but in a list?
And even if you specify a number, you get a list of your requested project as well as all projects.

I feel like this is trying to do too much? I've not gone through all the code so maybe there's a good reason for this, but seems odd to me at this stage.

EDIT:
I think I understand the reasoning here. Because (normally) you're updating a project as well as the aggregate at the same time.
Perhaps this could be renamed to `get_project_and_aggregate`, return a `tuple[ProjectStats, ProjectStats]`, and remove the defaulting to -1?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+
+            # Filter to valid CUIs.
+            valid_cuis = [
+                cui
+                for cui in cuis
+                if isinstance(cui, str)
+                and self.filters.check_filters(cui)
+            ]
+            if valid_cuis:
+                gold_anns.append({
```
mart-r, line 268: Would be great to have this format defined somewhere, e.g a `TypedDict`.

medcat-v2/medcat/utils/training_utils.py
```diff
@@ -88,9 +88,20 @@ def predict(doc: MutableDocument) -> list[MutableEntity]:
...
+                ent = tokens2entity(tkns, doc)
+            except ValueError:
+                while not tkns:
+                    # If no tokens found, try expanding the 
+                    # range by 1 character on each side
+                    start = max(0, start - 1)
+                    end = end + 1
+                    tkns = doc.get_tokens(start, end)
+                ent = tokens2entity(tkns, doc)
+                ent.id = -1000
```
mart-r: Would be great to have the `-1000` a constant somewhere (e.g `UNTOKENIZABLE_ENTITY_ID = -1000`). That way we could reuse that everywhere without fear of mistyping and with knowledge of what that means.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+                continue
+
+            extracted.append({
+                'start': ent.base.start_char_index,
+                'end': ent.base.end_char_index,
+                'cui': ent.cui,
+                'text': ent.base.text,
+                'confidence': float(ent.context_similarity),
+                'raw': ent,
+                'no_tokens': 1 if ent.id == -1000 else 0,
```
mart-r: Refer to the comment in `training_utils` where this is set on making it a constant .

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        apply_filters: bool = True,
+    ) -> list[dict]:
+        """Extract relevant info from predicted entities."""
+        extracted = []
+
+        for ent in predictions:
+            if apply_filters and not self.filters.check_filters(ent.cui):
+                continue
+
+            extracted.append({
```
mart-r, line 290: Again, a specified format would be great.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+                    linking=linking,
+                )
+                for i in range(num_projects)
+            },
+        )
+    
+class StatsCalculator:
+    """Calculates statistics for entity linking."""
+
+    BUCKET_FULL = 'full'
```
mart-r: Again, probably want these as an `Enum`.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        project_index: int,
+        mode: str
+    ) -> None:
+        """Count gold annotations for a project and all-projects aggregate."""
+        for project_stats in self.stats.get_projects(project_index):
+            mode_stats = project_stats.get_mode(mode)
+            if mode_stats is None:
+                continue
+            state = mode_stats.stats
+            if mode == self.BUCKET_NER:
```
mart-r, line 316: Not quite clear why this hack is needed here.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+                    state.cui_gold_counts.get(cui, 0)
+                    + 1
+                )
+        
+    def _record_tp(self, state: RawStats, gold: dict, pred: dict) -> None:
+        """Record a true positive."""
+        cui = pred['cui']
+        state.tp += 1
+        state.cui_tp[cui] = state.cui_tp.get(cui, 0) + 1
+        if cui not in state.examples['tp']:
```
mart-r, line 339: This only reports one example per concept (for each state)?
I think previously it would do all examples, no?

Is this a deliebarate change or an accidental one?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+                'start': pred['start'],
+                'confidence': pred['confidence']
+            })
+
+    def _record_fn(self, state: RawStats, gold: dict) -> None:
+        """Record a false negative."""
+        cui = gold['cui']
+        state.fn += 1
+        state.cui_fn[cui] = state.cui_fn.get(cui, 0) + 1
+        if cui not in state.examples['fn']:
```
mart-r, line 355: Same question as above - there will now only be one example per cui.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+                'acceptable_cuis': gold['cuis'],
+                'start': gold['start']
+            })
+        
+    def _record_fp(self, state: RawStats, pred: dict) -> None:
+        """Record a false positive."""
+        cui = pred['cui']
+        state.fp += 1
+        state.cui_fp[cui] = state.cui_fp.get(cui, 0) + 1
+        if cui not in state.examples['fp']:
```
mart-r, line 369: Same as above two.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        
+    def _score_annotations(self, 
+                           gold_anns: list[dict], 
+                           pred_anns: list[dict],
+                           project_index: int, 
+                           mode: str, 
+                           filter_fp_by_cui: bool = True) -> None:
+        # Track which predictions have been matched
+        matched_preds: set[int] = set()
+        all_projects_state = self.stats.all_projects.get_mode(mode)
```
mart-r: Perhaps should use `self.stats.get_projects` and use the list?
Seems like that's why the method was created?
Would need to get get the mode-sepcific stats for each project, of course. But then the subsequent things would iterate over the list rather than doing each one by one.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
-        anns_norm_neg: list[tuple[int, str]] = []
-        anns_examples: list[dict] = []
-        anns_norm_cui: list[str] = []
+                # False Negative
+                self._record_fn(all_projects_state.stats, gold)
+                self._record_fn(project_state.stats, gold)
+
+        # Phase 2: Remaining predictions are False Positives
+        for idx, pred in enumerate(pred_anns):
+            if idx not in matched_preds:
```
mart-r: [NAG]
Maybe invert out to lower indentation?
i.e `if idx in matched_preds: continue` and `if filter_fp_by_cui and not self.filters.check_filters(pred['cui']): continue`.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
-                if ann.get("validated", True):
-                    # This is used to test was someone annotating for this
-                    # CUI in this document
-                    anns_norm_cui.append(cui)
-                    self.cui_counts[cui] = self.cui_counts.get(cui, 0) + 1
-        return anns_norm, anns_norm_neg, anns_examples, anns_norm_cui
-
-    def finalise_report(self, epoch: int, do_print: bool = True):
-        """Finalise the report / metrics.
+            chars = set(range(start, end))
```
mart-r: [NAG]
Perhaps rename to `char_nums` or `char_idxs`? I've not got far enough to see the usage of this, but my guess is that it's for IoU. But when I read this line of code I was like "this is a set of integers not a set of characters as the variable name would suggest".

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        sklearn implementation, and it is also more memory efficient.
+        
+        Testing shows same metrics, and halving computation speed.
+        """
 
-        This prints out the overall metrics and calculates per CUI metrics.
+        tp = len(gold_chars & pred_chars)
+        fp = len(pred_chars - gold_chars)
+        fn = len(gold_chars - pred_chars)
+        tn = document_length - tp - fp - fn
```
mart-r, line 530: Is this a standard proxy for true negatives in the context of cohen kappa?
In principle, the true negatives here effectively infinite, as I'm sure you know.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        # cohen_k = sum of document-level macro Kappas
+        #         -> divide by number of documents
+        # cui_iou[CUI] = sum of per-document IoU for that CUI
+        #              -> divide by number of documents containing that CUI
+        # cui_giou[CUI] = sum of per-document GIoU for that CUI
+        #               -> divide by number of documents containing that CUI in gold
+        # cui_cohen_k[CUI] = sum of per-document CUI-specific Kappa
+        #                 -> divide by number of documents where the CUI is evaluated
+        """
+        state = self.stats.projects[project_index].get_mode(mode)
```
mart-r: Again, perhaps `self.stats.get_projects(project_index)` and a list?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+
+            state.stats.cui_cohen_k[cui].append(cohen_k)
+            all_project_state.stats.cui_cohen_k[cui].append(cohen_k)
+
+        # Average the per-CUI IoUs rather than merging character sets.
+        # This preserves CUI identity.
+        if doc_cui_ious:
+            doc_iou = sum(doc_cui_ious) / len(doc_cui_ious)
+        else:
+            doc_iou = 1.0
```
mart-r: Perhaps make this a default-valued keyword argument in case it makes sense to set it to 0 in some case?

To me, defaulting to IoU of 1 if there's no cuis provided or predicted feels like not necessarily expected result.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+
+        state.stats.iou_sum += doc_iou
+        all_project_state.stats.iou_sum += doc_iou
+
+        # Only gold CUIs contribute to GIoU, so we average over those.
+        if doc_cui_gious:
+            doc_giou = sum(doc_cui_gious) / len(doc_cui_gious)
+        else:
+            doc_giou = 1.0
```
mart-r: Same as above

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+
+        denominator = 1 - pe
+
+        if denominator == 0:
+            # Perfect agreement or no variation
+            return 1.0
+
+        return (po - pe) / denominator
+    
+    def _score_character_annotations(self,
```
mart-r: Perhaps there's a way to split this into two or three different methods?

I feel like it could look something like:
```python
def _calculate_document_cui_scores(self, *args):# didn't look at the specific arguments
     # the calculation logic,
     return doc_cui_ious, doc_cui_gious, doc_cui_kappas
def _update_project_stats(self, projects: list, doc_cui_ious, doc_cui_gious, doc_cui_kappas) -> None:
     # updates the projects' state
     pass
def _score_character_annotations(self, **):
     # keep doc string and all, of course
     my_projects = [proj.get_mode(mode) for proj in self.stats.get_projects(project_index)]
     doc_cui_ious, doc_cui_gious, doc_cui_kappas = self._calculate_document_cui_scores(my_projects, *args)
     self._update_project_stats(my_projects, doc_cui_ious, doc_cui_gious, doc_cui_kappas)
```

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+            )
+        
+    def process_project(self, project: MedCATTrainerExportProject,
+                        project_index: int,
+                        entity_getter: Callable[[str], list[MutableEntity]],
+                        mode: str,
+                        calculate_ner_performance: bool = False,
+                        use_project_filters: bool = False,
+                        extra_cui_filter: set[str] | None = None
+                        ) -> None:
```
mart-r, line 770: This would most certainly need a doc string.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        if doc is None:
+            return []
+        return doc.linked_ents
+
+    def process_export(self, cat: CAT, export: MedCATTrainerExport,
+                       mode: str,
+                       calculate_ner_performance: bool = False,
+                       use_project_filters: bool = False,
+                       extra_cui_filter: set[str] | None = None,
+                       filter_before_disamb: bool = False) -> None:
```
mart-r, line 809: This needs doc string.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        if filter_before_disamb:
+            cat.config.components.linking.filter_before_disamb = True
```
mart-r, line 822: This makes the change permanently, not only for the stats building time.
Perhaps use `with medcat.utils.config_utils.temp_changed_config` here?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+            return info.get('preferred_name') or list(info['names'])[0]
+        return cui
+    
+    def _safe_mean(self, values):
+        return sum(values) / len(values) if values else 0.0
+    
+    def compute_metrics(
+        self,
+        mode: str,
+        project_index: int = -1,
```
mart-r: What is the purpose of the -1 default?
Looks like that'll cause only the aggregate to be counted?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        """Get preferred name for CUI."""
+        info = self.cui2info.get(cui)
+        if info:
+            return info.get('preferred_name') or list(info['names'])[0]
+        return cui
+    
+    def _safe_mean(self, values):
+        return sum(values) / len(values) if values else 0.0
+    
+    def compute_metrics(
```
mart-r: Perhaps we can split this up as well?
I.e have this iterate over the projects, call another method for the preparation, and then set the metrics.
Something like:
```python
def _prepare_metricS(self, *args):
    # do the work
    return overall, per_cui
def compute_metrics(self, *args):
    for project_stats in self.stats.get_projects(project_index):
        overall, per_cui = self._prepare_metrics()
        mode_stats.metrics = Metrics(
                overall=overall,
                per_cui={
                    cui: CUIMetrics(**metrics)
                    for cui, metrics in per_cui.items()
                },
            )
```

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+    # these 3 functions are just copied from previous, 
+    # they get nice names for concepts
     def _empty(self, cui: str) -> CUIInfo:
         return get_new_cui_info(
             cui=cui, preferred_name=cui, names=set((cui, )))
 
     def _get_or_empty(self, cui: str) -> CUIInfo:
         return self.cui2info.get(cui, self._empty(cui))
-
+
```
mart-r: Whitespace added?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
 
     def _get_or_empty(self, cui: str) -> CUIInfo:
         return self.cui2info.get(cui, self._empty(cui))
-
+            
     def _get_pref_name(self, cui: str) -> str:
         info = self._get_or_empty(cui)
         return info['preferred_name'] or list(info['names'])[0]
+    
+    def print_stats(self,
```
mart-r, line 993: Since we're redoing this entire thing thoroughly, perhaps allow (and use) a stream in the method signature here?
That way we'd make it easier to capture the information (either in code or in specific files for logging).

I'd just say allow for a `stream: SupportsWrite[str] | None = None` keyword argument and pass that on to `print(*, file=stream)`.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+
+def get_stats(cat: CAT, 
+                  data: MedCATTrainerExport,
+                  epoch: int = 0,
+                  use_project_filters: bool = False,
+                  use_overlaps: bool = False,
+                  ner_performance: bool = False,
+                  linking_performance: bool = False,
+                  extra_cui_filter: Optional[set[str]] = None,
+                  do_print: bool = True,) -> "StatsCalculator":
```
mart-r: This is a breaking change in terms of the return type.

The problem is that we don't know whether or what is using our software somewhere downstream.
And as such, I'd be extremely reluctant in making a drastic change like this here. You can see the effects in the fact that the tutorials initially failed and needed to be patched.

For reference, this might break something UCLH folks are doing with the MiADE (recently updated (or mid update) to v2) or CogStack ModelServe (not sure whether they've full updated or which version of medcat they're using in production, but I know they did do a v2 update and were using stuff like `get_stats`).

I would prefer that the old signature remain (at least for now). I.e you'd unwrap the output like you've done in the tutorials or in `kfold` stats.
And this new stuff would be in another method, `get_stats_new`, `get_stats2`, or something like that.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,443 +1,981 @@
...
+        with dataset_aware_component(cat, CoreComponentType.ner, data):
+            calculator.process_export(
+                cat,
+                data,
+                mode=StatsCalculator.BUCKET_LINKING,
+                use_project_filters=use_project_filters,
+                extra_cui_filter=extra_cui_filter,
+            )
+
+
```
mart-r, line 1120: Perhaps stuff below here (and above `do_print`) should be a method on `StatsCalculator`? I.e `StatsCalculator.compute_all_metrics`
It should already have the necessary options. And would make this method a little neater.

mart-r (changes requested):
Overall, I think this is definitely a step in the right direction!
The setup looks good, the output seems easy to use.
And there's still the option to get the same sort of output.

I don't really think we can introduce a breaking change in the manner that you're doing it here due to unkown downstream effects.

So I'd say this thing (i.e the new returned object) needs to be in its own method and the `get_stats` needs to use this and unwrap the output (like you've done in various bits).

There's a few nagging things.
But also a few things that I think would need to change.
A few structures I'd like to be defined more rigidly (rather than just `dict` or predefined strings).
A few bits where I feel like we could easily split out the longer methods into smaller ones.
And then there's a matter of documentation in a few places.
And then one place where I asked for a feature for print output stream.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done!

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: See `StatsCalculator.legacy_stats`.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done!

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: yeah done. Just split it out to:

```
    def get_project_stats(self, project_index: int) -> ProjectStats:

    def get_aggregate_stats(self) -> ProjectStats:
```

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Sorry not sure exactly what needs a typed dict? the annotations or the valid_cuis?

medcat-v2/medcat/utils/training_utils.py
adam-sutton-1992: I've put it in `tokens.py`. I think its best there?

adam-sutton-1992: And of course, propagated it in stats.py

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Okay I think it's typed dicts for annotations and predictions. I'll give that a go and see what you think :D

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: 😏 mypy 😏

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Yes apologies.

I saw this in your `my_stats`. I erroneously thought it was only used there so remove it. When it came to testing I see you wanted this examples dict so I put it back from memory. I thought examples was only one per concept for tp, fp, fn. 

Just checking now I've missread it. Oops!

From a design perspective I think having all examples is a bit excessive. But it was there previously so I'll revert to that.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: I'm not sure how you think it's infinite, and yes my reading of it is that it's a proxy based on the fact that these are binary cohen kappas. It's either detected the entity or hasn't. In the case of multiple cuis it would be much more complex than this. 

I.e. a string with values labels [0,0,0,0,0,0,1,1,1,1,1,0,0,0,0,0,0,2,2,2,2,0,0,0,0,0,3,3,3,3,3,0,0,0,0,0,0] would be more complex than [0,0,0,0,0,0,1,1,1,1,1,0,0,0,0,0,0,1,1,1,1,1,0,0,0,0,0,1,1,1,1,1,0,0,0,0,0,0]. 

If you've got 8k characters, 200 of them are tps, 400 are fps, 300 are fns, then you've got 7,1k that you've either are neither labelled nor predicted to be entities and are the true negatives.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: I've set both to 0. I think in both either 1 or 0 is fine as they will be performances relative to each other. I chose 1 because I feel that is _is_ a perfect prediction in the case that happens (however rare it would be that you get into this edge case).

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: I've given it a go.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: I've wrapped `get_stats` around `get_stats_calculator`. Get Stats calculator will return the entire object. Get Stats will do as previous.

medcat-v2/medcat/stats/stats.py
mart-r: The format of the contents of `gold_anns`.
I.e 
```
class GoldenAnnotation(TypedDict):
    start: int
    end: int
    cuis: list[str]
    cui: str
    text: str
    raw: str
```
Don't need to use the constructor for it, but if you type the output, the type checkers (like `mypy`) can check that the stuff being accessed actually exists.

I.e have the return type here `list[GoldenAnnotation]` so everything down the road knows the format.

medcat-v2/medcat/stats/stats.py
mart-r: I agree - in principle I'd expect an example to be a subset of all...
Perhaps we can (outside the scope of this) optionally limit this. Because currently that's (by a large margin) the largest part of the output. So if somoene want to print this out (or save it) would be great if we could NOT have all the data represented in the examples section...

medcat-v2/medcat/stats/stats.py
mart-r: I get it now. This is on a per character basis. So in this case it does make sense that.

My assumption was that it's still on a per entity basis (like the classic metrics), in which case any length of substring could be a possible entity (and most would thus always be true negatives).

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done!

medcat-v2/medcat/stats/stats.py
mart-r: Do you still need the mapping here? Can't you just use the `.value` of the enum?

medcat-v2/tests/stats/test_kfold.py
```diff
@@ -215,6 +215,9 @@ def setUpClass(cls) -> None:
 
     def setUp(self) -> None:
         super().setUp()
+        # return (self.fps, self.fns, self.tps,
```
mart-r, line 218: Maybe we don't need the comments to be added here? Probably left over when you had the breaking change?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,309 +1,984 @@
...
+    def _count_gold_annotations(
+        self,
+        gold_anns: list[GoldAnnotation],
+        project_index: int,
+        mode: MetricMode
+    ) -> None:
+        """Count gold annotations for a project and all-projects aggregate."""
+        project_stats = self.stats.get_project_stats(project_index)
+        aggregate_stats = self.stats.get_aggregate_stats()
+        for project_stats in (project_stats, aggregate_stats):
```
mart-r, line 311: Supernag: the shadowing of `project_stats` (that is originally the specific project's stats rather than here being either that or the aggregate) is not ideal.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,309 +1,984 @@
...
+                    linking=linking,
+                )
+                for i in range(num_projects)
+            },
+        )
+    
+class StatsCalculator:
+    """Calculates statistics for entity linking."""
+
+    BUCKET_FULL = MetricMode.FULL
```
mart-r, line 210: NAG: Do we need to reference them here ? Can't we just use `MetricMode.NER` and/or `MetricMode.LINKING` directly?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,309 +1,984 @@
...
+        
+    def _score_annotations(self, 
+                           gold_anns: list[GoldAnnotation], 
+                           pred_anns: list[PredictedAnnotation],
+                           project_index: int, 
+                           mode: MetricMode, 
+                           filter_fp_by_cui: bool = True) -> None:
+        # Track which predictions have been matched
+        matched_preds: set[int] = set()
+        aggregate_stats = self.stats.get_aggregate_stats()
```
mart-r, line 432: NAG[Proj+Aggr]: I find it annoying that you're having to get both the project-specific stats and the aggregate states and work on them separately. I feel like it's bound to lead us to a situation where only one is changed by accident.

Perhaps using a list here (`project_and_aggregate = [all_projects_state, project_state]`) and iterating over this when updating?
Or using a composite that updates both? Though this would be more work + code to maintain.

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,309 +1,984 @@
...
+                gold_chars,
+                pred_chars,
+                doc_length,
+            )
+
+        return per_cui_ious, per_cui_gious, per_cui_kappas
+
+    def _update_project_stats(
+        self,
+        project_state: ModeStats,
```
mart-r, line 612: NAG[Proj+Aggr]: Same here, perhaps a list then?

medcat-v2/medcat/stats/stats.py
```diff
@@ -1,309 +1,984 @@
...
+        # cohen_k = sum of document-level macro Kappas
+        #         -> divide by number of documents
+        # cui_iou[CUI] = sum of per-document IoU for that CUI
+        #              -> divide by number of documents containing that CUI
+        # cui_giou[CUI] = sum of per-document GIoU for that CUI
+        #               -> divide by number of documents containing that CUI in gold
+        # cui_cohen_k[CUI] = sum of per-document CUI-specific Kappa
+        #                 -> divide by number of documents where the CUI is evaluated
+        """
+        aggregate_stats = self.stats.get_aggregate_stats()
```
mart-r, line 689: NAG[Proj+Aggr]: I guess this is where it starts.

medcat-v2/medcat/stats/stats.py
```diff
@@ -314,130 +989,164 @@ def _get_or_empty(self, cui: str) -> CUIInfo:
...
-    @classmethod
-    def from_cat(cls, cat: CAT,
-                 use_project_filters: bool = False,
-                 use_overlaps: bool = False,
-                 #  use_cui_doc_limit: bool = False,
-                 #  use_groups: bool = False,
-                 extra_cui_filter: Optional[set[str]] = None
-                 ) -> 'StatsBuilder':
-        """Get the stats builder from a model pack and some extra information.
+def get_stats_calculator(cat: CAT,
```
mart-r, line 1082: This could use a doc string.
I'd like to include why this might be better (i.e how you get granular access to the data), but also include the fact that the calculator has already done its job (the calculations) and there is no need to call an extra method to do that or something like that.

medcat-v2/medcat/stats/stats.py
```diff
@@ -314,130 +989,164 @@ def _get_or_empty(self, cui: str) -> CUIInfo:
...
+    calculator.compute_all_metrics(ner_performance, linking_performance)
+    
+    if do_print:
+        to_print = calculator.stats.all_projects.get_mode(StatsCalculator.BUCKET_FULL)
+        if to_print is None:
+            raise ValueError("No statistics available for the full pipeline mode.")
+        calculator.print_stats(epoch, to_print)
+    return calculator
+
+def get_stats(cat: CAT,
```
mart-r, line 1128: This also does still need a doc string. Doesn't need to be the same as it had before, but something at least.

mart-r (changes requested):
A few doc strings I'd like to see, plus removal of added comments from `test_kfold.py`.

And the project + aggregate issue that's still present. Would be nice to clean that up, but I don't think it's high priority. Can easily leave as is.

The rest is more or less just nagging.

mart-r:
NOTE:
On top of the above, now needs resynced with master due to folder change in #607

medcat-v2/tests/stats/test_kfold.py
adam-sutton-1992: Oops

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Fixed - sorry didn't notice that.

medcat-v2/medcat/stats/stats.py
adam-sutton-1992: Done, as a list.

medcat/medcat/stats/stats.py
```diff
@@ -1,309 +1,984 @@
...
+        # cohen_k = sum of document-level macro Kappas
+        #         -> divide by number of documents
+        # cui_iou[CUI] = sum of per-document IoU for that CUI
+        #              -> divide by number of documents containing that CUI
+        # cui_giou[CUI] = sum of per-document GIoU for that CUI
+        #               -> divide by number of documents containing that CUI in gold
+        # cui_cohen_k[CUI] = sum of per-document CUI-specific Kappa
+        #                 -> divide by number of documents where the CUI is evaluated
+        """
+        aggregate_stats = self.stats.get_aggregate_stats()
```
adam-sutton-1992: And then changed to here, and removed from update_project_stats

medcat/medcat/stats/stats.py
```diff
@@ -314,130 +989,164 @@ def _get_or_empty(self, cui: str) -> CUIInfo:
...
-    @classmethod
-    def from_cat(cls, cat: CAT,
-                 use_project_filters: bool = False,
-                 use_overlaps: bool = False,
-                 #  use_cui_doc_limit: bool = False,
-                 #  use_groups: bool = False,
-                 extra_cui_filter: Optional[set[str]] = None
-                 ) -> 'StatsBuilder':
-        """Get the stats builder from a model pack and some extra information.
+def get_stats_calculator(cat: CAT,
```
adam-sutton-1992: Done

medcat/medcat/stats/stats.py
```diff
@@ -314,130 +989,164 @@ def _get_or_empty(self, cui: str) -> CUIInfo:
...
+    calculator.compute_all_metrics(ner_performance, linking_performance)
+    
+    if do_print:
+        to_print = calculator.stats.all_projects.get_mode(StatsCalculator.BUCKET_FULL)
+        if to_print is None:
+            raise ValueError("No statistics available for the full pipeline mode.")
+        calculator.print_stats(epoch, to_print)
+    return calculator
+
+def get_stats(cat: CAT,
```
adam-sutton-1992: Done, I also removed the options for linking performance and ner performance, as they're never used in legacy stats.

mart-r (approved):
Looking great!
Thanks for the work!

PR #607, opened 2026-08-18
docs(medcat): Move away from v2 prefix

mart-r:
This PR moves away from `medcat-v2` prefix to just `medcat`.

While the change itself is not super massive, it will touch a lot of files that had links or other references to the v2 path.
The most immediate changes are:
- `medcat-v2` folder renamed to `medcat`
- `medcat-v2-tutorials` folder renamed to `medcat-v2-tutorials`

But there's a lot of things in many different parts (especially with local install / test against core lib state) that also needed changed.

PS:
This PR will trigger almost all the workflows (didn't check, but just a guess - it targets most of the larger ones at least). So I wouldn't be surprised if something is broken somewhere. Especially if it hits stuff that we've not touched in a while.

EDIT:
The docs build initially failed since the config on their side says to explicitly look for `medcat-v2/readthedocs.yaml` and that can no longer be found. But I did change it (tempoarily) to the new `medcat/readthedocs.yaml`. If/when this PR gets merged, the change would need to be made permanent.
In any case, here's the link to the docs:
https://cogstack-platform-toolkit--607.org.readthedocs.build/projects/nlp/en/607/

EDIT2:
~~For some reason a bunch of the links in the docs are not working.
Not entirely sure why that is.
As far as I can tell, I've done everything in the same way that it was done in the previous `medcat[-v2]/mkdocs.yaml`.
But I'm sure I just missed something obvious.~~

EDIT3:
I've since resolved the above issue. Turned out the conversion I had done was all good. But the issue was in how some plugin was handling the folder name of `medcat/medcat` (previously `medcat-v2/medcat`). Because of both folders having the same name, some plugins didn't work as expected.
So the fix was to add a hook to fix the links.
The reason these needed to be fixed in the first place is because they're in the `README.md` which is pulled into `docs/index.md`. However, the links are relative to the `medcat` folder rather than the `medcat/docs` folder. Previously this was correctly fixed automatically, but now it was not due to the confusion with the folder names.
The other caveat is that the hook had to change `README.md` during `index.md` read since the inclusion of it within `index.md` happens by another plugin at a later stage.


medcat/docs/hooks.py
```diff
@@ -0,0 +1,16 @@
+import re
+from pathlib import Path
+
+
+def on_page_markdown(markdown, page, config, files):
```
alhendrickson: I'd propose that now is the time to invert the README/index.md thing as mentioned before.

So currently, the mkdocs build has to pull in the README.file as a snippet. I'd want to flip this round, so that the README.md file effectively just says "See [here] for documentation". This is the pattern medcat-trainer is using (https://github.com/CogStack/cogstack-nlp/blob/main/medcat-trainer/README.md). We make index.md the contents (and specifically, the html page should be the documentation not the markdown )

The reason I think now is the time - on docs.cogstack.org I'm actually importing the docs using mkdocs-multirepo, so this hooks.py file wont get used at all and would really not be possible. I think any further complexity here is going to get really difficult, when we can just simplify it like trainer has

medcat/docs/v1_conversion.md
```diff
@@ -0,0 +1,20 @@
+# Converting a v1 model to v2 format
```
alhendrickson, line 1: This should also get added to mkdocs.yaml

alhendrickson (changes requested):

medcat/docs/v1_info.md
```diff
@@ -0,0 +1,9 @@
+## Why MedCAT v2?
+
+MedCAT v2 is a comprehensive refactor designed to improve modularity, flexibility, and maintainability. The core library is now lightweight, with optional extras (spaCy tokenization, MetaCAT, DeID, RelCAT$
```
alhendrickson:
```suggestion
MedCAT v2 is a comprehensive refactor designed to improve modularity, flexibility, and maintainability. The core library is now lightweight, with optional extras (spaCy tokenization, MetaCAT, DeID, RelCAT)
```

Typo on relcat to prove you're not an LLM

medcat/docs/hooks.py
mart-r: Makes sense. Will be done in a bit.

medcat/docs/v1_conversion.md
mart-r: It's there already, no?
https://cogstack-platform-toolkit--607.org.readthedocs.build/projects/nlp/en/607/v1_conversion/
It's linked from v1_info:
https://cogstack-platform-toolkit--607.org.readthedocs.build/projects/nlp/en/607/v1_info/

alhendrickson (approved):
Thanks for the changes - lgtm

PR #613, opened 2026-08-28
fix(medcat-den): CPDT-171 Fix den model load api for core lib 2.9

mart-r:
This PR fixs MedCAT den workflow by updating the API in line with medcat==2.9.0 where a new (optional) argument was added to `CAT.load_model_pack`.


alhendrickson (approved):
lgtm

PR #619, opened 2026-09-15
Mcat service rel cat improvement

vladd-bit:
- Added relations to medcat service output.
- Added relations to medcat-v2 pipeline output.
- Corrected cdb loading for RelCAT module
- Added concept types to relation output
- Added component settings to m-catservice for future use.

Draft.


medcat-tutorials/notebooks/introductory/relcat/1._Supervised_Training_Relation_Extraction.ipynb
```diff
@@ -101,7 +101,7 @@
    ],
    "source": [
     "# Install medcat\n",
-    "! pip install \"medcat[spacy,rel-cat,meta-cat]~=2.4.0\" # NOTE: VERSION-STRING"
+    "! pip install \"medcat[spacy,rel-cat,meta-cat]~=2.9.0\" # NOTE: VERSION-STRING"
```
mart-r: During smoke tests, this is run within the notebook before anything else.

So as much as the CI already installs the up to date source-based (i.e current state) version, this will always just pull the recent release from PyPI and overwrite anyway.

That's why we needed the hack that you've removed.

.github/workflows/medcat-v2-tutorials_main.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           python-version: "3.11"
           cache: 'pip'
 
-      - name: Update install targets in notebooks
+      - name: Skip notebook MedCAT installs (installed by CI)
```
mart-r: Why the name change?

For reference, this was meant to not only do the install, but also change the install targets in notebooks because otherwise at smoke test time the notebooks (that require specific extras) would install the latest PyPI version anyway as part of the first code block.

medcat-service/medcat_service/nlp_processor/medcat_processor.py
```diff
@@ -135,6 +136,57 @@ def process_entities(self, entities, *args, **kwargs):
...
+    def _get_entities(self, text: str, disabled_components: tuple[str, ...]):
+        if not disabled_components:
+            return self.cat.get_entities(text)
+
+        cat = cast(CAT, self.cat)
+        pipeline = cat.pipe
+        disabled_names = {self._normalise_component_name(name) for name in disabled_components}
+        matched_names: set[str] = set()
+        doc = pipeline.tokenizer(text)
+        # MedCAT does not currently expose this as a public get_entities option.
```
mart-r, line 175: I think something like this might make a bit more sense:
```python
for component in self._filter_pipeline_components(pipeline.iter_all_components(), disabled_names, matched_names):
    doc = component(doc)
```
I.e avoid the two separate filter operations and separate loops, plus avoid using pipe internals (`._*`).

mart-r (changes requested):
One minor thing for the disabled components bit to avoid using internals.

Is there a reason to have a disabled components use case? Is there a specific use case in mind? Perhaps we could move that to the core lib if that's the case? Just thinking out loud. I'll be happy with the change I proposed.

vladd-bit:
> One minor thing for the disabled components bit to avoid using internals.
> 
> Is there a reason to have a disabled components use case? Is there a specific use case in mind? Perhaps we could move that to the core lib if that's the case? Just thinking out loud. I'll be happy with the change I proposed.

The idea would be to have something like http://host/disabled_components=rel_cat, for example, the use case here being that some people might not really want to run rel_cat over the text/annotations since it takes ages to compute without filters  for just even 1 document (at lest on cpu).  I still need to add an option to do relations only for example if only rel_cat is enabled and everything else is disabled in the near future, also need to change acceptable payloads.

mart-r (approved):
Looks good!

PR #620, opened 2026-09-21
minor stats fixes to printing and NER only measurement

adam-sutton-1992:
Hihi - 

Just some minor changes I found while using the new stats:

* `get_stats` and `get_stats_calculator` now print appropriate metrics for legacy and new stats. The spacing is cleaned up slightly too.
* The NER only pipeline would assign all labels as "__NER__" - which was a bit ugly / confusing. I was going to set it to "0"  or "1" but instead opted for "DETECTED_ENTITY" which I think is the most descriptive.


adam-sutton-1992:
How does my own ruff not spot that over indenting... honestly...

mart-r (approved):
Looks good

